PR #10, opened 2018-02-28
Setup automated testing on Jenkins CI

dalg24:
@mkstoyanov this is a work in progress.  Do not merge yet.

### DIFF
--- a/docker/jenkins/build.sh
+++ b/docker/jenkins/build.sh
@@ -8,7 +8,6 @@ cmake \
   -D CMAKE_CXX_FLAGS="-Wall -Wextra" \
   -D CMAKE_CXX_COMPILER=g++ \
   -D PYTHON_EXECUTABLE=/usr/bin/python3 \
-  -D Tasmanian_ENABLE_MATLAB=OFF \
   ..
 make -j${NPROC}
 ctest -j${NPROC} --no-compress-output -T Test
--- a/docker/jenkins/build.sh
+++ b/docker/jenkins/build.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e
+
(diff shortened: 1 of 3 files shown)


dalg24:
@mkstoyanov this ready for your review

docker/jenkins/build.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -e
+
+[ -d build ] && rm -rf build
+mkdir build && cd build
+cmake \
+  -D CMAKE_CXX_FLAGS="-Wall -Wextra" \
+  -D CMAKE_CXX_COMPILER=g++ \
+  -D PYTHON_EXECUTABLE=/usr/bin/python3 \
+  -D Tasmanian_ENABLE_MATLAB=OFF \
```
mkstoyanov: I fixed that, MATLAB is disabled by default and this line is not needed any more.

docker/jenkins/build.sh
```diff
@@ -0,0 +1,14 @@
...
+  ..
+make -j${NPROC}
+ctest -j${NPROC} --no-compress-output -T Test
```
mkstoyanov: Do not use "ctest -j***". By default, Tasmanian uses OpenMP and -j can lead to more threads than cores and end up with slower runtime. Also, there are only 4 ctests (if you enable Fortran), multi-threading will probably not help here.

docker/jenkins/build.sh
dalg24: Will do.

docker/jenkins/build.sh
dalg24: With CMake you can set properties on a test and specify how many process slots are required.

PR #11, opened 2018-03-01
Specify number of threads to use in tests

dalg24:
This should address your comment about using `-j<num_procs>` when running the tests in #10 

Basically we can specify how many process slots a test requires. I chose 4 as the default. I can change it if you like. Note that I only listed `Tasgrid` and `Tasdream`. Let me know if I should include others.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -850,6 +850,13 @@ endif()
 if (Tasmanian_ENABLE_FORTRAN)
     add_test(Fortran fortester)
 endif()
+include(ProcessorCount)
+ProcessorCount(NUM_PROCS)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -850,6 +850,11 @@ endif()
 if (Tasmanian_ENABLE_FORTRAN)
     add_test(Fortran fortester)
 endif()
+set(Tasmanian_TESTS_OMP_NUM_THREADS 4 CACHE STRING "Number of OpenMP threads to use in tests (default is 4)")
```
mkstoyanov: Let's talk about the purpose of this tomorrow. We should meet and discuss many other things going forward.

dalg24:
@mkstoyanov I would typically squash the commits marked with `[fixup]` into the 1st one when the review is done, just before the merge.

PR #13, opened 2018-03-02
warnings are disabled by default

mkstoyanov:
warnings are disabled by default, unless Tasmanian_ENABLE_DEVELOPMENT_DEFAULTS is used


mkstoyanov:
Let's see if I did this right this time, mostly regarding the way that I made the pull-request

CMakeLists.txt
```diff
@@ -429,7 +429,7 @@ else()
 endif()
 
 # for development purposes, compile the exmaples together with everything else
-if (DEFINED Tasmanian_DEVELOPMENT_EXAMPLES AND Tasmanian_DEVELOPMENT_EXAMPLES)
+if (DEFINED Tasmanian_ENABLE_DEVELOPMENT_DEFAULTS AND Tasmanian_ENABLE_DEVELOPMENT_DEFAULTS)
```
dalg24: `if (Tasmanian_ENABLE_DEVELOPMENT_DEFAULTS)` already does what you want. It evaluate to false if undefined.

CMakeLists.txt
```diff
@@ -669,20 +669,17 @@ endif()
 # Compiler specific flags: Intel hasn't been tested in a while
 ########################################################################
 if (NOT DEFINED Tasmanian_CXX_FLAGS)
-    if (${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -Wall -Wextra -Wshadow -Wuninitialized -Wstrict-aliasing -pedantic")
-        #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -ffast-math -mtune=native -Wall -Wextra -Wshadow -pedantic") # -Wno-unused-parameter
+    if ((${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU") OR (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang"))
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
```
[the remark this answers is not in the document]
dalg24: Also I checked on my laptop with a dummy CMakeLists.txt and `-O3` seems to be the default.
```
//Flags used by the compiler during debug builds.
CMAKE_CXX_FLAGS_DEBUG:STRING=-g

//Flags used by the compiler during release builds for minimum
// size.
CMAKE_CXX_FLAGS_MINSIZEREL:STRING=-Os -DNDEBUG

//Flags used by the compiler during release builds.
CMAKE_CXX_FLAGS_RELEASE:STRING=-O3 -DNDEBUG

//Flags used by the compiler during release builds with debug info.
CMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING=-O2 -g -DNDEBUG
```

CMakeLists.txt
dalg24: Also I am getting confused by the body of that if-statement. Is the purpose of that variable only to enable the examples? If so all it should do is add the executable and link. Or even better, set a variable `Tasmanian_ENABLE_EXAMPLES` to true and later conditionally add the directory `Examples/` and let it have it's own `CMakeLists.txt` that adds the executable, place the headers in include search path and link libraries as needed.

CMakeLists.txt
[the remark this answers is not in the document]
mkstoyanov: There is not reason not to use -O3 in Debug mode. If I "Debug", then I want to debug the -O3. In some cases, classes compile different in "Debug" mode to allow more rigorous error checking, but Tasmanian has no such classes. The only difference between Debug and Release is the debug symbols. I may have to change that in the future, but it is OK for now. I see it is redundant for Release, I can fix that in the future.

CMakeLists.txt
mkstoyanov: Examples have their own CMakeLists, which points to the installed version of Tasmanian. However, when I develop the examples, I need to constantly recompile and I didn't want to go through the sequence "CMake Tasmanian -> Install -> CMake Examples", I wanted to have the examples compiled in the Tasmanian build folder. This is nothing but convenience, and it does not affect the user in any way.
I used to have "Tasmanian_DEVELOPMENT_EXAMPLES", but I changed that to "Tasmanian_ENABLE_DEVELOPMENT_DEFAULTS" which covers both the examples and the CXX flags, the development section enables -Wall -Wextra -pedantic and so on. It is something for you and me to use, but no the user.

mkstoyanov: The cmake system needs an overhaul, but I have an ECP milestone due at the end of the month and I have to make a large Tasmanian presentation in April. I will go over cmake once more, break it into sub-files for the different sections, and change all of the Defined statements. Right now, the statement above was changes with "search-replace".

PR #16, opened 2018-03-09
Set up multi job on jenkins-ci

dalg24:


mkstoyanov:
Looks good, no we can do more than one test and it seems the clang5 test passes as well. If you're ready, you can merge it yourself.

PR #22, opened 2018-04-05
Add CUDA to automated testing

### LINKED ISSUE #15: Enable CUDA in the CI
Perform CI testing with CUDA, right now the docker image does not support cuda.

Note that cmake needs:
-D Tasmanian_ENABLE_CUBLAS:BOOL=ON (set by default)
-D CUDA_TOOLKIT_ROOT_DIR:PATH=<path-to-cuda> (if FindCUDA doesn't find it automatically)

Off-line testing is currently done with 7.5, 8.0, 9.0, and 9.1. We can pick whatever is the easiest "default" by NVIDIA or latest version 9.1. It matters little for the CI.

We have to make sure Tasmanian build with cuda support and cmake did not automatically disable it. For example:
[[ -z `tasgrid -v | grep cublas` ]] || { echo "Tasmanian was not compiled with cublas support"; exit 1;}

dalg24:
Fix #15

### DIFF
--- a/docker/Dockerfile_nvidia-cuda9.0-ubuntu16.04
+++ b/docker/Dockerfile_nvidia-cuda9.0-ubuntu16.04
@@ -1,6 +1,7 @@
 FROM nvidia/cuda:9.0-devel-ubuntu16.04
 
 RUN apt-get update && apt-get install --no-install-recommends -y \
+        libopenblas-dev \
         python \
         python-numpy \
         cmake \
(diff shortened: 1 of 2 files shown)


mkstoyanov:
Looks like it's working. You can merge it, if you're done. Thanks!
Could we also add BLAS (e.g., libopenblas-dev) to either of the builds. Doesn't really matter which one.

dalg24:
I added BLAS to the new image with CUDA 9.0

PR #23, opened 2018-04-06
enable cuda tests and add comments

mkstoyanov:


mkstoyanov:
The alternative solution is to manually add -std=c++11 to CUDA_NVCC_FLAGS, but then I have to remove "-pedantic"

Config/CMakeIncludes/compiler_flags_extras.cmake
```diff
@@ -1,11 +1,16 @@
...
-if (Tasmanian_ENABLE_CXX_11 OR Tasmanian_ENABLE_MPI OR
-    Tasmanian_ENABLE_CUDA OR Tasmanian_ENABLE_CUBLAS)
-    foreach(Tasmanian_loop_target ${Tasmanian_target_list})
-        set_property(TARGET ${Tasmanian_loop_target} PROPERTY CXX_STANDARD 11)
-    endforeach()
+# if Tasmanian_ENABLE_CXX_11 is defined and set to off
+# then we do not enable c++11
+if ((NOT DEFINED Tasmanian_ENABLE_CXX_11) OR Tasmanian_ENABLE_CXX_11)
+    if (Tasmanian_ENABLE_CXX_11 OR Tasmanian_ENABLE_MPI OR
+        Tasmanian_ENABLE_CUDA OR Tasmanian_ENABLE_CUBLAS)
```
dalg24: No!

mkstoyanov: You don't see the logic is changed by adding "if ((NOT DEFINED Tasmanian_ENABLE_CXX_11) OR Tasmanian_ENABLE_CXX_11)" or you want to force c++11.
I'm making a release next week. After the release, I will make changes and will require c++11 and newer cmake and other newer things.

dalg24: Over complicated

mkstoyanov: I agree, but months ago I released version 5.0 which had the (over complicated) options to force enable/disable C++-11. The simple thing is to require C++-11, but I don't want to push new requirements for version 5.1, which is supposed to be just a polished 5.0 with just a few extras.

After I release 5.1, c++-11 will be mandatory, cmake will have to be at least 3.5, and so on.

Have a nice weekend and thanks for all the help!

PR #24, opened 2018-04-07
Cuda updates

mkstoyanov:
Added fail safety mechanism to ensure kernels are not called with wrong order. Added GPU acceleration for local polynomials and evaluateFast(). There is an option to force sparse backend acceleration algorithm. Cleaned Cuda warnings and function calls.


mkstoyanov:
That's all the "new" features for 5.1

PR #28, opened 2018-04-13
added tested cuda version to the manual, and gpu use in matlab

mkstoyanov:


dalg24:
Maybe consider using git-lfs for binary files.

PR #33, opened 2018-04-24
Simple defines

mkstoyanov:
* streamlined cmake variables, e.g., Tasmanian_ENABLE_CUDA is used in both cmake and c++ without the need for the intermediary TASMANIAN_CUDA; change also covers MPI and CUBLAS
* only python files are auto-generated for simple GNU-Make build, TasmanianConfig.hpp and tsgGetPaths.m are simple enough to maintain manually
* cached test threads and gpu variables and the MATLAB work folder; the behavior of the build engine has not changed, this makes things cleaner when using the cmake gui
* fixed some typos


CMakeLists.txt
```diff
@@ -82,6 +80,10 @@ option(Tasmanian_ENABLE_CUDA "Enable Nvidia GPU acceleration using CUDA kernels
 option(Tasmanian_STATIC_LIBRARY "Compile static library and the executables with static linkage (Tasmanian)" ON)
 option(Tasmanian_SHARED_LIBRARY "Compile dynamic library (Tasmanian)" ON)
 
+set(Tasmanian_TESTS_GPU_ID "-1" CACHE STRING "specify GPU ID for testing, -1 means running tests on all visiable devices")
```
dalg24: typo `s/visiable/visible`

Config/AltBuildSystems/TasmanianConfig.hpp
```diff
@@ -31,9 +31,9 @@
 #ifndef __TASMANIAN_CONFIG_HPP
```
dalg24: Should this file only be generated at the time of release?

Config/AltBuildSystems/TasmanianSG.py
```diff
@@ -39,7 +39,7 @@
     bTsgPlotting = False
     tsgPlot = []
 
-__version__ = "5.1"
+__version__ = "6.0"
```
dalg24: Should this be coming from the `C/C++`?

Config/TasmanianConfig.in.hpp
```diff
@@ -45,9 +45,9 @@
...
 ////////////////////////////////////////////////////////////////////////
 
-#cmakedefine TASMANIAN_CPU_BLAS
-#cmakedefine TASMANIAN_CUBLAS
-#cmakedefine TASMANIAN_CUDA
-#cmakedefine TASMANIAN_MPI
+#cmakedefine Tasmanian_ENABLE_BLAS
+#cmakedefine Tasmanian_ENABLE_CUBLAS
+#cmakedefine Tasmanian_ENABLE_CUDA
+#cmakedefine Tasmanian_ENABLE_MPI
```
dalg24: Sounds good.

Config/AltBuildSystems/TasmanianSG.py
mkstoyanov: All version info comes from the same line in cmake, there is also regression test that ensures that those will match.

Config/TasmanianConfig.in.hpp
mkstoyanov: Should have done this from the beginning.

PR #34, opened 2018-04-24
Simple defines

mkstoyanov:
* streamlined cmake variables, e.g., Tasmanian_ENABLE_CUDA is used in both cmake and c++ without the need for the intermediary TASMANIAN_CUDA; change also covers MPI and CUBLAS
* only python files are auto-generated for simple GNU-Make build, TasmanianConfig.hpp and tsgGetPaths.m are simple enough to maintain manually
* cached test threads and gpu variables and the MATLAB work folder; the behavior of the build engine has not changed, this makes things cleaner when using the cmake gui
* fixed some typos


dalg24:
Why didn't you just force pushed onto `mkstoyanov:simple_defines` and merge #33 ?

mkstoyanov:
I wanted to "rebase -i" to squash small "[fix]" commits, but in the past I messed up a few rebases so now I rebase in a separate branch to prevent a mistake (rw stands for rewritten).

I should check if this isn't the proper way to do this.

dalg24:
>  I rebase in a separate branch to prevent a mistake

Fair enough.  Then you could have done (assuming that the remote is called `mkstoyanov`)
```bash
$ git push mkstoyanov simple_defines_rw:simple_defines
```

mkstoyanov:
`$ git push mkstoyanov simple_defines_rw:simple_defines`

Makes sense, I'll do that next time.

dalg24:
You will probably need to add the `-f` option to force the push.

PR #36, opened 2018-04-30
Custom random generator for DREAM

mkstoyanov:
* added custom random number generator for all DREAM tests; this will remove the dependence on the compiler provided rand() calls and remove compiler dependence for the tests
* DREAM uses the overwritten random number generator to select the chains for computing increments; there is no place in the code left, where compiler provided rand() would be used when a custom rng is provided

### DIFF
--- a/DREAM/tasdreamExternalTests.cpp
+++ b/DREAM/tasdreamExternalTests.cpp
@@ -48,7 +48,8 @@ double TestRNG::getSample01() const{
     return (double)(((double) s) / 2097165.0);
 }
 
-ExternalTester::ExternalTester(int num_monte_carlo) : num_mc(num_monte_carlo){
+ExternalTester::ExternalTester(int num_monte_carlo) : num_mc(num_monte_carlo), rngseed(-1){
+    // rngseed = -1 indicates to use the seed coded for each test
     #ifdef _TASMANIAN_WINDOWS_
     srand(15);
     #else
@@ -58,7 +59,7 @@ ExternalTester::ExternalTester(int num_monte_carlo) : num_mc(num_monte_carlo){
     verbose = false;
 }
 ExternalTester::~ExternalTester(){}
-void ExternalTester::resetRandomSeed(){ srand(time(0)); }
+void ExternalTester::resetRandomSeed(){ rngseed = time(0); }
 
 void ExternalTester::setVerbose(bool new_verbose){ verbose = new_verbose; }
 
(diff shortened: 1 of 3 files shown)


DREAM/tasdreamExternalTests.cpp
```diff
@@ -38,7 +38,18 @@ using std::cerr;
...
-ExternalTester::ExternalTester(int num_monte_carlo) : num_mc(num_monte_carlo){
+TestRNG::TestRNG(int seed): s(seed % 2097165){}
+TestRNG::~TestRNG(){}
+
+double TestRNG::getSample01() const{
+    s *= 511;
+    s += 127;
+    s %= 2097165;
+    return (double)(((double) s) / 2097165.0);
+}
```
dalg24: This is the linear congruential algorithm isn't it?  Why don't you use what is implemented in the standard header `<random>`?

dalg24: C.f. http://en.cppreference.com/w/cpp/numeric/random

mkstoyanov: Yes, this is a rather basic algorithm, but it is sufficient. I was using <random>, but that was the main problem with clang5, and then happened again when I was using Visual C++. GCC, different versions of Clang (e.g., 3.8, vs 4.0 vs 5.0), and MS Visual Studio, all use different random number generators, meaning that the same seed gives different sequences of samples for the different compilers. I'm guessing ICC and PGI will be also different. DREAM is a random sampler and tests have to be done statistically, e.g., with 95% or 99% confidence. I have 5 - 6 tests, if each passes at the 95% then I have 25% failure rate based on the random seed. Due to the differences in compilers, I cannot simply fix the seed ... with a custom compiler, I don't have to worry about that.

This is done only for the tests, by default, the library uses <random> and the user can opt to specify their own random number generator.

dalg24:
```C++
// I would just use the standard minstd_rand engine.
// But you can always replace with `std::linear_congruential_engine<std::uint_fast32_t, 512, 127, 2097165>` if you like.
// Pass a seed to the constructor if you don't like the default.
std::minstd_rand generator();

// Then I assume you were trying to get a uniform distribution over [-1, 1]
std::uniform_real_distribution<double> distribution(-1., 1.);

// Then do
for (int i; ...)
  sample[i] = distribution(generator);  
```

mkstoyanov: I see your point, C++11 does have standard that should be compiler independent. However, using this won't save on code unless I also change the way the core distribution is assigned to the DREAM and CorePDF classes. It is a design decision to consider, I don't want to block the user from specifying something potentially more exotic than what is in the C++11 standard.

PR #38, opened 2018-06-07
Cmake modernize

mkstoyanov:
* split the build system into sub-directories, overall the options and behavior remains mostly the same, but the cmake organization is different
* most of dependencies are now specified per-target following modern cmake style (whenever possible, see code comments about OpenMP)
* fixed the OpenMP bug, if using cmake 3.9 or newer library dependence is passed correctly
* now it is possible to build the examples and the library with different compilers (cmake 3.9 or newer only)
* cmake 3.5 is still supported, but OpenMP requires a global flag and identical compiler to link to the libraries (workaround is possible but it is not clean)
* it is possible to include Tasmanian into a larger project with the add_subdirectory() command, exports for install may still use some cleaning
* install folder structure is changed, cmake is moved to lib/Tasmanian/, everything that doesn't go into bin or lib goes into share/Tasmanian. In principle it is possible to install Tasmanian in /usr/local, but it is not recommended
* added lots of comments on problematic and questionable places, could use more comments


CMakeLists.txt
```diff
@@ -184,13 +229,18 @@ if (NOT USE_XSDK_DEFAULTS)
     endforeach()
     if (Tasmanian_ENABLE_MATLAB)
         message(" -D Tasmanian_MATLAB_WORK_FOLDER:PATH=${Tasmanian_MATLAB_WORK_FOLDER}")
-        message(" pre-install MATLAB folder: addpath('${CMAKE_BINARY_DIR}/matlab/')")
+        message(" pre-install MATLAB folder: addpath('${CMAKE_CURRENT_BINARY_DIR}/MATLAB/matlab/')")
```
dalg24: Why did you change to `CMAKE_CURRENT_BINARY_DIR`?  Why did you add a "MATLAB" directory?

Config/CMakeIncludes/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
...
+set(Tasmanian_post_install "${Tasmanian_post_install}                  see CMakeLists.txt in ${CMAKE_INSTALL_PREFIX}/examples\n\n")
+
+set(Tasmanian_post_install "${Tasmanian_post_install}bash export commands (add to your environment, i.e., ~/.bashrc):\n\n")
+set(Tasmanian_post_install "${Tasmanian_post_install}export PATH=$PATH:${CMAKE_INSTALL_PREFIX}/bin/\n")
+set(Tasmanian_post_install "${Tasmanian_post_install}export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${CMAKE_INSTALL_PREFIX}/lib/\n")
+
+set(Tasmanian_post_install "${Tasmanian_post_install}above commands are writte in the following\n")
+set(Tasmanian_post_install "${Tasmanian_post_install}bash source files: ${CMAKE_INSTALL_PREFIX}/config/TasmanianENVsetup.sh\n")
+set(Tasmanian_post_install "${Tasmanian_post_install}cmake import:             ${CMAKE_INSTALL_PREFIX}/lib/Tasmanian/Tasmanian.cmake\n")
+set(Tasmanian_post_install "${Tasmanian_post_install}cmake directive:          include(\\\"${CMAKE_INSTALL_PREFIX}/lib/Tasmanian/Tasmanian.cmake\\\")\n")
```
dalg24: triple escape?  oh boy!

Config/CMakeIncludes/sanity_check_and_xsdk.cmake
```diff
@@ -174,20 +177,42 @@ if (Tasmanian_ENABLE_MPI)
...
-    # this is a bug in some versions of CUDA
-    if ((Tasmanian_ENABLE_CUBLAS OR Tasmanian_ENABLE_CUDA) AND ("${CUDA_VERSION_MAJOR}" LESS 8))
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_FORCE_INLINES")
+
+########################################################################
+# Compiler specific flags: Intel hasn't been tested in a while
+# Tasmanian_STRICT_OPTIONS=ON will prevent Tasmanian from setting
+# compiler flags, only the user provided flags will be used
+########################################################################
+if (NOT Tasmanian_STRICT_OPTIONS)
```
dalg24: I don't know how I feel about these, especially with default `Tasmanian_STRICT_OPTIONS` is `OFF`.

Config/CMakeIncludes/sanity_check_and_xsdk.cmake
```diff
@@ -174,20 +177,42 @@ if (Tasmanian_ENABLE_MPI)
...
+# the extra variables here allow the user to circumvent the problem by
+# including additional directories, also check SparseGrids/CMakeLists.txt
+# comment about Tasmanian_EXTRA_LIBRARIES
+########################################################################
+if (DEFINED Tasmanian_EXTRA_INCLUDE_DIRS)
+    include_directories(${Tasmanian_EXTRA_INCLUDE_DIRS})
+endif()
+
+if (DEFINED Tasmanian_EXTRA_LINK_DIRS)
+    link_directories(${Tasmanian_EXTRA_LINK_DIRS})
```
dalg24: Same comment as for `include_directories` here.  Why did you choose to do this over `target_link_directories()`?

DREAM/tasdreamExternalTests.cpp
```diff
@@ -278,7 +273,7 @@ bool ExternalTester::testGamma1D(){
 }
 
 bool ExternalTester::testGaussian2D(){
-    int s = (rngseed == -1) ? 12 : rngseed;
+    int s = (rngseed == -1) ? 25 : rngseed;
```
dalg24: Did you mean to change that in this PR??

SparseGrids/CMakeLists.txt
```diff
@@ -0,0 +1,266 @@
+cmake_minimum_required(VERSION 3.5)
```
dalg24: Usually only one require at the root of the project.

CMakeLists.txt
```diff
@@ -58,119 +60,162 @@ message(STATUS "This is a development version of Tasmanian")
...
+# 5. Add the "make test_install" command
+# 6. Install top level files
+#    - cmake export file
+#    - master config file
+#    - CMakeLists.txt for the examples (include OpenMP hack if needed)
+#    - shell (bash) source file that defines paths
+# 7. Print final messages
+########################################################################
+
+option(USE_XSDK_DEFAULTS        "Enable xSDK compatibility"              OFF)
```
dalg24: Is this variable name imposed by XSDK?

CMakeLists.txt
```diff
@@ -2,8 +2,8 @@ cmake_minimum_required(VERSION 3.5)
 
 cmake_policy(VERSION 3.5)
 project(Tasmanian VERSION 6.0.0 LANGUAGES CXX)
-set(CMAKE_CXX_STANDARD 11)
 message(STATUS "This is a development version of Tasmanian")
+set(Tasmanian_license "BSD 3-Clause with UT-Battelle disclaimer") # used in some headers and python modules (only human readable)
```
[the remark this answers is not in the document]
mkstoyanov: Variables with upper case are user defined (e.g., Tasmanian_ENABLE_OPENMP), variables with lower case are internal for Tasmanian (e.g., Tasmanian_license or Tasmanian_python_install_path). The user doesn't get to change the licence.

Will fix the typo.

CMakeLists.txt
```diff
@@ -20,19 +20,19 @@ message(STATUS "This is a development version of Tasmanian")
 #    -D Tasmanian_STATIC_LIBRARY:BOOL=ON
 #    -D Tasmanian_MATLAB_WORK_FOLDER:PATH=""
 #
-# Additional options:
+# Additional options: (if default test behavior is not desired)
 #    -D Tasmanian_TESTS_OMP_NUM_THREADS:INT="sets OpenMP number of threads"
 #    -D Tasmanian_TESTS_GPU_ID:INT="specifies which GPU to use for testing"
-#    -D Tasmanian_EXTRA_CXX_FLAGS:STRING="appends to the default flags"
```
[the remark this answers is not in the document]
mkstoyanov: Already removed, just forgot to remove the comment.

CMakeLists.txt
```diff
@@ -20,19 +20,19 @@ message(STATUS "This is a development version of Tasmanian")
...
-#    -D Tasmanian_EXTRA_INCLUDE_DIRS:PATH="appends more include paths"
+#
+# Extra options: (allow for workarounds on bugs and exotic systems)
 #    -D Tasmanian_EXTRA_LIBRARIES:STRING="appends more link libraries"
+#    -D Tasmanian_EXTRA_INCLUDE_DIRS:PATH="appends more include paths"
 #    -D Tasmanian_EXTRA_LINK_DIRS:PATH="appends more link paths"
```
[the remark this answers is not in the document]
mkstoyanov: None that I could find. I could hack this by adding -I<dir> directives to the CXX flags, but not sure if I can do it with libraries. Also, I do want this done per-target, which relates to a comment below.

CMakeLists.txt
```diff
@@ -41,6 +41,8 @@ message(STATUS "This is a development version of Tasmanian")
 #    -D XSDK_ENABLE_CUDA:BOOL=OFF
 #
 # Alternatives to find_package()
+# (each find_package() defines a set of variables, if the sue manually
```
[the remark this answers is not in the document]
mkstoyanov: s/sue/user, will fix

CMakeLists.txt
```diff
@@ -58,119 +60,162 @@ message(STATUS "This is a development version of Tasmanian")
...
 option(Tasmanian_ENABLE_FORTRAN "Enable Fortran interface for Tasmanian" OFF)
-option(Tasmanian_ENABLE_CUBLAS "Enable Nvidia GPU acceleration using cuBLAS for Tasmanian" ON)
-option(Tasmanian_ENABLE_CUDA "Enable Nvidia GPU acceleration using CUDA kernels for Tasmanian" OFF)
+option(Tasmanian_ENABLE_CUBLAS  "Enable Nvidia GPU acceleration using cuBLAS for Tasmanian"       ON)
+option(Tasmanian_ENABLE_CUDA    "Enable Nvidia GPU acceleration using CUDA kernels for Tasmanian" OFF)
 option(Tasmanian_STATIC_LIBRARY "Compile static library and the executables with static linkage (Tasmanian)" ON)
 option(Tasmanian_SHARED_LIBRARY "Compile dynamic library (Tasmanian)" ON)
 
-set(Tasmanian_TESTS_GPU_ID "-1" CACHE STRING "specify GPU ID for testing, -1 means running tests on all visible devices")
+# threat those similar to options, but give values other than ON/OFF
```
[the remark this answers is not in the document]
mkstoyanov: Thanks!

CMakeLists.txt
```diff
@@ -58,119 +60,162 @@ message(STATUS "This is a development version of Tasmanian")
...
+set(Tasmanian_TESTS_GPU_ID          "-1" CACHE STRING "specify GPU ID for testing, -1 means running tests on all visible devices")
```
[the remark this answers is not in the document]
mkstoyanov: `-1` means hit all visible GPUs, otherwise pick one GPU. I should make this INT instead of a string.

I could accept `0,1` but it is too much work.

CMakeLists.txt
```diff
@@ -58,119 +60,162 @@ message(STATUS "This is a development version of Tasmanian")
...
-include(${CMAKE_SOURCE_DIR}/Config/CMakeIncludes/source_file_list.cmake)
+include("${CMAKE_CURRENT_SOURCE_DIR}/Config/CMakeIncludes/sanity_check_and_xsdk.cmake")
 
 
 ########################################################################
-# Sanity check, xSDK compatibility, and find_package() calls
+# Core project configuration
 ########################################################################
-include(${CMAKE_SOURCE_DIR}/Config/CMakeIncludes/sanity_check_and_xsdk.cmake)
+enable_testing() # call in one place, each subfolder gives additional tests
```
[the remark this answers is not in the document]
mkstoyanov: Currently, the tests are included in the cli wrapper and compile anyway. The python tests do not compile, i.e., they take no time. If the user doesn't want tests, then just skip the `make test`command.

CMakeLists.txt
```diff
@@ -58,119 +60,162 @@ message(STATUS "This is a development version of Tasmanian")
...
 
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/Config/TasmanianConfig.in.hpp"
+               "${CMAKE_CURRENT_BINARY_DIR}/configured/TasmanianConfig.hpp")
+if(${CMAKE_VERSION} VERSION_LESS "3.9.0") # check the correct version here
+    # older versions of cmake do not add incude directories for cuda targets
+    include_directories("${CMAKE_CURRENT_BINARY_DIR}/configured/")
+endif()
```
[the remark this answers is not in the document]
mkstoyanov: `target_include_directories(<configured>)` is called for all appropriate targets, but cmake prior to 3.9 doesn't pass those to `nvcc`. We require cmake `3.5` to avoid being aggressive, so older cmake files require an extra include directive to ensure those get passed to nvcc.

CMakeLists.txt
```diff
@@ -58,119 +60,162 @@ message(STATUS "This is a development version of Tasmanian")
...
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/Config/TasmanianConfig.in.hpp"
+               "${CMAKE_CURRENT_BINARY_DIR}/configured/TasmanianConfig.hpp")
+if(${CMAKE_VERSION} VERSION_LESS "3.9.0") # check the correct version here
+    # older versions of cmake do not add incude directories for cuda targets
+    include_directories("${CMAKE_CURRENT_BINARY_DIR}/configured/")
+endif() 
 
 ########################################################################
 # Setup targets
+# CXX subdirs have to come first, SparseGrids must preceed DREAM
```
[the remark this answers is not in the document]
mkstoyanov: Yep!

CMakeLists.txt
```diff
@@ -58,119 +60,162 @@ message(STATUS "This is a development version of Tasmanian")
...
+# The order of the interfaces is not important
+# each sub-directory creates a set of targets and links those to
+# existing targets (hence the order is important)
```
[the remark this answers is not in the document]
mkstoyanov: CXX is the core; Python, Fortran and MATLAB are just interfaces. Thus the core CXX must come first, and Sparse Grids must come before DREAM. Once the core is in place, the interfaces can be in any order. Will clarify.

CMakeLists.txt
mkstoyanov: All paths are relative now, so Tasmanian can be added with `add_subdir()`.

The MATLAB folder contain the scripts so those can be used during build and before install. This is done for testing and sometimes convenience. Double MATLAB may be redundant, I'll think about it.

Config/CMakeIncludes/CMakeLists.txt
mkstoyanov: Yep, I want `" "` to appear in code, so I had to escape it, but it gets processed twice, one when the variable is set, once when the post-install message is printed. Hence, I need to escape the escape character.

Config/CMakeIncludes/sanity_check_and_xsdk.cmake
mkstoyanov: This is done for low level users that don't want to read messages. We can default to `ON` and set it in the install script.

There are almost no flags, just bug work-around. The strict options affects what happens if `find_package(BLAS)` or `find_package(CUDA)` fail.

XSDK sets this to on.

Config/CMakeIncludes/sanity_check_and_xsdk.cmake
mkstoyanov: I should have done that. I will do it.

SparseGrids/CMakeLists.txt
mkstoyanov: OK, will remove the extra ones.

DREAM/tasdreamExternalTests.cpp
mkstoyanov: I think this was related to windows testing and the way windows flags are handled, 12 was failing on windows because of differences in the new build system.

CMakeLists.txt
mkstoyanov: Yes: 

`https://xsdk.info/policies/`

check the second large link: 

`xSDK community installation policies: GNU Autoconf and CMake options [click here] – version 0.3.0, Nov 6, 2017`

CMakeLists.txt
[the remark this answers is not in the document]
dalg24: I actually did not intend to change the spelling.  And I am OK with your explanation.

Config/CMakeIncludes/sanity_check_and_xsdk.cmake
mkstoyanov: I will use `target_include_directories()`, but `link_directories()` cannot be specified per-target. Those should rarely be necessary, but can give you a fighting chance when dealing with bugs or exotic setup.

DREAM/tasdreamExternalTests.cpp
mkstoyanov: Actually, this was related to the build flags in the simple GNU Make build engine. Still, this is part of the build engine modifications.

PR #39, opened 2018-06-22
Fourier basis

zbmorrow:
This branch implements sparse Fourier interpolation. The syntax of `GridFourier::makeGrid` is

```c++
void GridFourier::makeGrid(int num_dimensions, int num_outputs, int depth, const int *anisotropic_weights = 0, const int *level_limits = 0)
```
similarly to the syntax of `GridGlobal::makeGrid`. The other user-facing commands (`loadNeededPoints`, `evaluate`, `evaluateBatch`, etc.) behave the same as in the `GridGlobal` class.

The one-dimensional Fourier grid on the canonical domain [0,1] has 3^_l_ (_l_>=0) equally spaced points

and the set of one-dimensional basis functions on level _l_ is

where . Note that the sets of nodes and basis functions are nested with respect to _l_.

In order to construct the interpolation coefficients, we must perform a discrete Fourier transform of the y-data (in each output direction) on the constituent tensor-product grids. This is done once, when the function values are loaded at the front end, and is decoupled from the evaluation of the complex exponential basis functions, which must be done for each evaluation point. The DFT is implemented in `TasmanianFourierTransform::dft` within `tsgLinearSolvers.cpp`.


SparseGrids/tsgGridFourier.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include "tsgIndexManipulator.hpp"
+#include "tsgLinearSolvers.hpp"
+#include "tsgOneDimensionalWrapper.hpp"
+#include "tsgGridCore.hpp"
+
+#include "tsgAcceleratedDataStructures.hpp"
+
+namespace TasGrid{
+
+class GridFourier{
```
mkstoyanov: * must inherit from `BaseCanonicalGrid`
* must also add multiple `evaluate()` functions, you can make those fallback to basic `evaluate()` and `evaluateBatch()`

SparseGrids/tsgGridFourier.hpp
```diff
@@ -0,0 +1,91 @@
...
+    const double* getLoadedValues() const;
+    
+    double* getLoadedPoints() const;
+    void getLoadedPoints(double *x) const;
+    double* getNeededPoints() const;
+    void getNeededPoints(double *x) const;
+    double* getPoints() const;
+    void getPoints(double *x) const; // returns the loaded points unless no points are loaded, then returns the needed points
+
+	void calculateFourierCoefficients();
```
mkstoyanov: * tab for indentation, use spaces

SparseGrids/tsgGridFourier.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+    std::complex<double>* getBasisFunctions(const double x[]) const;
```
mkstoyanov: * shouldn't those be `protected`?

SparseGrids/tsgGridFourier.hpp
```diff
@@ -0,0 +1,91 @@
...
+    void getBasisFunctions(const double x[], std::complex<double> weights[]) const;
+
+    void evaluate(const double x[], double y[]) const;
+    void evaluateBatch(const double x[], int num_x, double y[]) const;
+
+	void clearAccelerationData();
```
mkstoyanov: * check indentation across all files, there are many places where tabs are used instead of spaces

SparseGrids/tsgGridFourier.hpp
```diff
@@ -0,0 +1,91 @@
...
+    IndexSet *active_tensors;
+    int *active_weights;
+    IndexSet *points;
+    IndexSet *needed;
+    IndexSet *exponents;
+    
+	std::complex<double> *fourier_coefs;
+    int **exponent_refs;
+
+    int *max_levels; // for evaluation purposes, counts the maximum level in each direction (only counts tensors)
```
mkstoyanov: * is this used, `GridFourier::evaluate()` doesn't seem to need this.

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -167,6 +167,76 @@ void TasmanianTridiagonalSolver::decompose(int n, double d[], double e[], double
...
+    std::complex<double> i(0,1);
+
+
+    int k_idx[rank];
+    std::fill(k_idx, k_idx+rank, 0);
+    int k_restart_dim = 1;
+    std::complex<double>* exp_k = new std::complex<double>[rank];
+    std::fill(exp_k, exp_k+rank, 1);
+
+    for(int k=0; k<num_data; k++) {
```
mkstoyanov: * this is the $O(n^2)$ algorithm, right? If so, we need to think about this.

SparseGrids/tsgOneDimensionalWrapper.cpp
```diff
@@ -183,6 +183,8 @@ OneDimensionalWrapper::OneDimensionalWrapper(const OneDimensionalMeta *meta, int
         }else if ((rule == rule_mindelta) || (rule == rule_mindeltaodd)){
             GreedySequences greedy;
             unique = greedy.getMinDeltaNodes(meta->getNumPoints(max_level, rule));
+        }else if (rule==rule_fourier) {
+            unique = OneDimensionalNodes::getFourierNodes(max_level);
```
mkstoyanov: * wrapper is missing quadrature weights for integration

SparseGrids/tsgGridFourier.cpp
```diff
@@ -0,0 +1,365 @@
...
+    } else {
+        values = new StorageSet(num_outputs, needed->getNumIndexes());
+    }
+	
+}
+
+int* GridFourier::referenceExponents(const int levels[], const IndexSet *list) {
+	
+	// This is like IndexManipulator::referenceNestedPoints, but it references an ordered list of exponents for basis functions
+    int *num_points = new int[num_dimensions];
```
mkstoyanov: * memory leak here, this needs to be `delete[]`-ed

SparseGrids/tsgLinearSolvers.cpp
zbmorrow: Yes, the algorithm I wrote is $O(n^2)$ (one-time at the front end, but could still be bad for grids with large depth).

SparseGrids/tasgrid_main.cpp
```diff
@@ -156,6 +156,7 @@ int main(int argc, const char ** argv){
             if ((strcmp(argv[k],"global") == 0)) test = test_global;
             if ((strcmp(argv[k],"local") == 0)) test = test_local;
             if ((strcmp(argv[k],"wavelet") == 0)) test = test_wavelet;
+            if ((strcmp(argv[k],"fourier") == 0)) test = test_fourier;
```
mkstoyanov: * looks much better
* you need to add a test to `SparseGrids/CMakeLists.txt` with cmake command `add_test()`
* then we can test on the CI

```
add_test(SparseGridsFourier      tasgrid -test fourier)
```

zbmorrow: Done

mkstoyanov:
OK to test

zbmorrow:
Updates:

- Tests for Fourier rule, acceleration, and domain transform (refinement omitted for now)
- Tests incorporated into `tasgridExternalTests` (added a couple new functions to `tasgridTestFunctions`); tests also run upon install
- Added BLAS for complex matvec/matmat
- Implemented `getInterpolationWeights` for nodal interpolation (does DFT of complex exponentials on each tensor)
- Implemented quadrature rule
- Inherits from `BaseCanonicalGrid` class and is fully implemented in `TasmanianSparseGrid` class

Still needed:
- IO for binary/ASCII
- O(N log N) Fourier transform algorithm
- Is there a nicer way to re-index when reordering the data for the Fourier transform?

mkstoyanov:
@zbmorrow: check the comments, there is something wrong with the BLAS fall-back functions

mkstoyanov:
@zbmorrow, so far so good.  

* Next, we need to rebase to current master to fix the conflicts, then I'll go over the code again to see if I catch anything or if we have to massage the internals.
* We will work on more tests and performance afterwards.

SparseGrids/tasgridTestFunctions.cpp
```diff
@@ -80,6 +80,14 @@ TwoOneCosCos::TwoOneCosCos(){} TwoOneCosCos::~TwoOneCosCos(){} int TwoOneCosCos:
...
 void TwoOneCosCos::eval(const double x[], double y[]) const{ y[0] = cos(0.5 * M_PI * x[0]) * cos(0.5 * M_PI * x[1]); } void TwoOneCosCos::getIntegral(double y[]) const{ y[0] = 16.0 / (M_PI * M_PI); }
 
+TwoOneExpSinCos::TwoOneExpSinCos() {} TwoOneExpSinCos::~TwoOneExpSinCos() {} int TwoOneExpSinCos::getNumInputs() const{ return 2; } int TwoOneExpSinCos::getNumOutputs() const{ return 1; }
+const char* TwoOneExpSinCos::getDescription() const{ return "f(x,y) = exp(sin(2*pi*x) + cos(2*pi*y)) on [0,1]^2 "; }
+void TwoOneExpSinCos::eval(const double x[], double y[]) const{ y[0] = exp(sin(2*M_PI*x[0])+cos(2*M_PI*x[1])); } void TwoOneExpSinCos::getIntegral(double y[]) const{ y[0] = 1.6029228068079633; }
+
+TwoTwoSinCos::TwoTwoSinCos() {} TwoTwoSinCos::~TwoTwoSinCos() {} int TwoTwoSinCos::getNumInputs() const{ return 2; } int TwoTwoSinCos::getNumOutputs() const{ return 2; }
+const char* TwoTwoSinCos::getDescription() const{ return "f(x,y) = [sin(2*pi*x)cos(4*pi*y), cos(2*pi*x)]"; }
+void TwoTwoSinCos::eval(const double x[], double y[]) const{ y[0] = sin(2*M_PI*x[0])*cos(4*M_PI*x[1]); y[1]=cos(2*M_PI*x[0]); } void TwoTwoSinCos::getIntegral(double y[]) const{ y[0] = 0.0; y[1]=0.0; }
```
mkstoyanov: * `cos(2 * M_PI * x)` is too simple, try something that doesn't belong to the basis

SparseGrids/tasgridExternalTests.cpp
```diff
@@ -89,13 +91,14 @@ TestResults ExternalTester::getError(const BaseFunction *f, TasGrid::TasmanianSp
     int num_outputs = f->getNumOutputs();
     int num_points = grid->getNumPoints();
     if ((type == type_integration) || (type == type_nodal_interpolation)){
-        double *points = grid->getPoints(), *weights = 0;
+        double *points = grid->getPoints();
+        double *weights = new double[num_points];
```
mkstoyanov: Why assigning weights here? you are overwriting the pointer below.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -217,7 +217,7 @@ OneDimensionalMeta::~OneDimensionalMeta(){}
 const CustomTabulated* OneDimensionalMeta::getCustom() const{  return custom;  }
 
 int OneDimensionalMeta::getNumPoints(int level, TypeOneDRule rule) const{
-    int lcc;
+    int lcc, fourier_result;
```
mkstoyanov: You can reuse `lcc`, no need for another variable here.

SparseGrids/tsgGridFourier.cpp
```diff
@@ -0,0 +1,365 @@
...
+    for(int j=0; j<num_dimensions; j++){  num_points[j] = wrapper->getNumPoints(levels[j]); num_total *= num_points[j];  }
+
+    int* refs = new int[num_total];
+    int *p = new int[num_dimensions];
+
+    for(int i=0; i<num_total; i++){
+        int t = i;
+        for(int j=num_dimensions-1; j>=0; j--){
+            int tmp = t % num_points[j];
+            if(tmp <= (num_points[j]-1)/2) {
```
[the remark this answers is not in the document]
mkstoyanov: * Did you fix this, or am I reading something old?

SparseGrids/tsgGridFourier.cpp
```diff
@@ -0,0 +1,574 @@
...
+    
+    calculateFourierCoefficients();
+}
+
+
+
+double* GridFourier::getLoadedPoints() const{
+    if (points == 0) return 0;
+    int num_points = points->getNumIndexes();
+    double *x = new double[num_dimensions * num_points];
```
mkstoyanov: * no need for a variable
```
double *x = new double[num_dimensions * points->getNumIndexes()];
```

SparseGrids/tsgGridFourier.cpp
```diff
@@ -0,0 +1,574 @@
...
+
+double* GridFourier::getPoints() const{
+    return ((points == 0) ? getNeededPoints() : getLoadedPoints());
+}
+
+void GridFourier::getPoints(double *x) const{
+    if (points == 0){ getNeededPoints(x); }else{ getLoadedPoints(x); };
+}
+
+int GridFourier::convertIndexes(const int i, const int levels[]) const {
```
mkstoyanov: * we will have to fix the indexing and the fast algorithm later

SparseGrids/tsgGridFourier.cpp
```diff
@@ -0,0 +1,574 @@
...
+void GridFourier::evaluateFastGPUmagma(const double x[], double y[], std::ostream*) const{
+    evaluateFastCPUblas(x,y);
+}
+
+void GridFourier::evaluateBatchCPUblas(const double x[], int num_x, double y[]) const{
+    #ifdef Tasmanian_ENABLE_BLAS
+    int num_points = points->getNumIndexes();
+    std::complex<double> *y_tmp = new std::complex<double>[num_outputs * num_x];
+    std::complex<double> *weights = new std::complex<double>[num_points * num_x];
+    #pragma omp parallel for
```
mkstoyanov: * isn't this the same as: `evaluateHierarchicalFunctions()`

SparseGrids/tsgGridFourier.cpp
```diff
@@ -0,0 +1,574 @@
...
+            q[k] += w[i] * v[k];
+        }
+    }
+    delete[] w;
+}
+
+void GridFourier::evaluateHierarchicalFunctions(const double x[], int num_x, double y[]) const{
+    // y must be of size num_x * num_nodes * 2
+    int num_points = getNumPoints();
+    //#pragma omp parallel for
```
mkstoyanov: * Why skip OpenMP here?

SparseGrids/tsgGridFourier.cpp
```diff
@@ -0,0 +1,574 @@
...
+const double* GridFourier::getFourierCoefs() const{
+    double* fc = new double[2*getNumPoints()];
+    for(int i=0; i<exponents->getNumIndexes(); i++) {
+        fc[2*i] = fourier_coefs[i].real();
+        fc[2*i+1] = fourier_coefs[i].imag();
+    }
+    return fc;
+}
+
+bool GridFourier::blasEnabled() const {
```
mkstoyanov: * `TasmanianSparseGrids` class has `enabled/disabled` functions, what is the purpose of this function here?

SparseGrids/tsgGridGlobal.cpp
```diff
@@ -916,13 +916,13 @@ double* GridGlobal::computeSurpluses(int output, bool normalize) const{
         int qn = gg->getNumPoints();
         double *w = gg->getQuadratureWeights();
         double *x = gg->getPoints();
-        double *I = new double[qn];
+        double *I_quad = new double[qn];
```
mkstoyanov: * Is this change still needed?

SparseGrids/tsgGridFourier.hpp
```diff
@@ -0,0 +1,117 @@
+
+#ifndef __TASMANIAN_SPARSE_GRID_FOURIER_HPP
+#define __TASMANIAN_SPARSE_GRID_FOURIER_HPP
+
+#include <cstdlib>
+#include <math.h>
+#include <complex>
```
mkstoyanov: * Is it necessary to include the `<complex>` header here? 
* Can this be hidden in the `.cpp` implementation without propagating to the user?

SparseGrids/tsgHiddenExternals.hpp
```diff
@@ -30,6 +30,8 @@
 
 #ifndef __TASMANIAN_SPARSE_GRID_HIDDEN_INTERNALS_HPP
 #define __TASMANIAN_SPARSE_GRID_HIDDEN_INTERNALS_HPP
+#include <complex>
+using std::complex;
```
mkstoyanov: * *Never* force a namespace like that, all headers that include this would now work in the `std::complex` namespace
* if you need a variable from `std::complex`, then use that explicitly
```
using std::complex <some symbol>
```
* sometimes is it better to keep explicitly including from the namespace rather than contaminating your list of symbols

SparseGrids/tsgHiddenExternals.hpp
```diff
@@ -41,8 +43,10 @@ extern "C" double ddot_(const int *N, const double *x, const int *incx, const do
 extern "C" double dnrm2_(const int *N, const double *x, const int *incx);
 extern "C" void dgemv_(const char *transa, const int *M, const int *N, const double *alpha, const double *A, const int *lda, const double *x, const int *incx, const double *beta, const double *y, const int *incy);
 extern "C" void dtrsv_(const char *uplo, const char *trans, const char *diag, const int *N, const double *A, const int *lda, const double *x, const int *incx);
+extern "C" void zgemv_(const char *transa, const int *M, const int *N, const complex<double> *alpha, const complex<double> *A, const int *lda, const complex<double> *x, const int *incx, const complex<double> *beta, const complex<double> *y, const int *incy);
```
mkstoyanov: * isn't this the same as:
```
extern "C" ... const int *N, const double *alpha, ...
```
with the added note that alpha is 2 doubles

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -167,6 +167,74 @@ void TasmanianTridiagonalSolver::decompose(int n, double d[], double e[], double
...
+    // Fourier transform code by ZBM
+    // DOES NOT NORMALIZE the outgoing Fourier coefficients
+
+    if(direction != 1 && direction != -1) std::cerr << "ERROR: direction must be -1 (forward) or 1 (backward); value is " << direction << std::endl;
+    int num_data = 1;
+    for(int r=0; r<rank; r++) {
+        num_data *= n[r];
+    }
+
+    std::complex<double> i(0,1);
```
mkstoyanov: `i(0.0, 1.0)`

* With templates, you must be extra explicit when declaring thing, the compiler has to guess lots of stuff already and the chance for error is increased.

SparseGrids/TasmanianSparseGrid.cpp
```diff
@@ -220,6 +221,17 @@ void TasmanianSparseGrid::makeWaveletGrid(int dimensions, int outputs, int depth
         std::copy(level_limits, level_limits + dimensions, llimits);
     }
 }
+void TasmanianSparseGrid::makeFourierGrid(int dimensions, int outputs, int depth, TypeDepth type, const int* anisotropic_weights, const int* level_limits){
+    clear();
+    fourier = new GridFourier();
+    fourier->makeGrid(dimensions, outputs, depth, type, anisotropic_weights, level_limits);
+    base = fourier;
+    if(level_limits != 0) {
```
mkstoyanov: * The code must be formatter to be consistent with the rest (as well as self-consistent)
    * sometimes you leave a space between `) {`, do not
    * one space after `if (`
    * spaces around `=`, e.g., `i = 1`, there are exceptions here but only when trying to save space and make the code compact
    * there are places with multiple empty lines between functions, use one or zero lines
    * remove all trailing spaces and empty lines

SparseGrids/tasgridExternalTests.cpp
```diff
@@ -933,8 +936,8 @@ bool ExternalTester::testAllWavelet() const{
         //cout << setw(wfirst) << "Rules" << setw(wsecond) << TasGrid::OneDimensionalMeta::getIORuleString(rule_wavelet) << setw(wthird) << "Pass" << endl;
         cout << setw(wfirst) << "Rules" << setw(wsecond) << "wavelet" << setw(wthird) << "Pass" << endl;
     }else{
-        cout << setw(wfirst) << "Rule" << setw(wsecond) << TasGrid::OneDimensionalMeta::getIORuleString(rule_wavelet) << setw(wthird) << "FAIL" << endl; pass = false;
-    }{ TasGrid::TasmanianSparseGrid grid;
+        cout << setw(wfirst) << "Rules" << setw(wsecond) << TasGrid::OneDimensionalMeta::getIORuleString(rule_wavelet) << setw(wthird) << "FAIL" << endl; pass = false;
```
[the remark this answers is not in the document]
zbmorrow: It was a small thing I noticed. The line in the success part of the `if` statement has `Rules`. Based on the tests for global, sequence, etc., it looked as though when the tests for all rules of a given grid class pass, the program prints `Rules so-and-so Pass` or `Rules so-and-so FAIL`. I'll undo it in this branch since it's completely unrelated to the Fourier stuff.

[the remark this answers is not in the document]
mkstoyanov: * Global and Sequence relate to multiple `Rules`, hence each rule has a `Rule` pass/fail and the group has a group `Rules` pass/fail
* Wavelet rules have two variants, even though it is technically the same name
* Fourier has not variants (yet), but keep it the same as the rest

SparseGrids/tsgGridFourier.hpp
zbmorrow: When I move it to the `.cpp` and attempt to `make`, I get errors because the functions in the `.hpp` files have inputs of type `std::complex<typename>`.

mkstoyanov: * Yes, the question pertains to whether it is possible to redesign the header so that you avoid including `<complex>`

SparseGrids/tasgridExternalTests.cpp
zbmorrow: Reverted. I did this while trying to isolate some weird behavior, and I forgot to change it back.

SparseGrids/tsgGridFourier.cpp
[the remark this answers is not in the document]
zbmorrow: Huh, could have sworn I fixed this... I screwed up halfway through the rebase yesterday, which could be the problem. Fixed now at any rate.

SparseGrids/tsgGridGlobal.cpp
zbmorrow: Nope

SparseGrids/tsgGridFourier.cpp
zbmorrow: `evaluateHierarchicalFunctions()` gives the weights as an array of `double`s (real and imag parts), rather than complex numbers. Should I park this code in an overloaded function that gives the complex weights?

mkstoyanov: Here you have the exact same loop as in `evaluateHierarchicalFunctions()`, just call the function, why write the loop twice?

SparseGrids/TasmanianSparseGrid.cpp
zbmorrow: I believe I have taken care of items 1, 2, 4, and 5.  

For item 3, I did a reg exp search of (at least most of) the documents where I've made changes. The only instances I saw were in `for` loops. Was there a place in particular where I needed to fix the spaces around `=`?

The regexp I used was `[^( |=|(for\(int (a-z)))]=[^( |=)]`

mkstoyanov: The spaces around `=` go usually outside of the loop, I have to go back to find the spot where I saw it. Look over for code such as
```
base=fourier;
```

SparseGrids/tsgGridFourier.cpp
zbmorrow: Done

SparseGrids/TasmanianSparseGrid.cpp
zbmorrow: I think I found it in `tasgridTestFunctions.cpp`.

SparseGrids/tasgridWrapper.cpp
```diff
@@ -1048,6 +1097,8 @@ bool TasgridWrapper::executeCommand(){
         createLocalPolynomialGird();
     }else if (command == command_makewavelet){
         createWaveletGird();
+    }else if (command == command_makefourier){
+        createFourierGird();
```
mkstoyanov: * see earlier comment about merging this with `makeSequenceGrid()`

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -167,6 +167,68 @@ void TasmanianTridiagonalSolver::decompose(int n, double d[], double e[], double
...
+    // Fourier transform code by ZBM
+    // DOES NOT NORMALIZE the outgoing Fourier coefficients
+
+    if (direction != 1 && direction != -1) std::cerr << "ERROR: direction must be -1 (forward) or 1 (backward); value is " << direction << std::endl;
+    int num_data = 1;
+    for(int r=0; r<rank; r++){
+        num_data *= n[r];
+    }
+
+    std::complex<double> i(0.0, 1.0);
```
mkstoyanov: If the complex constant is defined in `std::complex::I`, then there is no need to redefine it here, just use it. I think it will work if you just say:
```
using std::complex::I;
```
and then use `I` afterwards.

zbmorrow: I get an error with `std::complex::I`. I looked on Stackexchange but couldn't find `sqrt(-1)` hardcoded somewhere as a constant within `std::complex` in C++11.

SparseGrids/tasgridWrapper.cpp
```diff
@@ -424,10 +463,10 @@ void TasgridWrapper::outputHierarchicalCoefficients() const{
     int num_p = grid->getNumPoints();
     int num_d = grid->getNumOutputs();
     if (outfilename != 0){
-        writeMatrix(outfilename, num_p, num_d, coeff, useASCII);
+        writeMatrix(outfilename, num_p, num_d, coeff, useASCII, grid->isFourier());
```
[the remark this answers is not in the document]
zbmorrow: Why not handle `printMatrix` the same as `writeMatrix`?
```
printMatrix(num_p, (grid->isFourier() ? 2 * num_d : num_d), coeff);
```

SparseGrids/tsgLinearSolvers.cpp
mkstoyanov: @zbmorrow I thought we had a problem with `I` being declared as `sqrt(-1)` and that's why you had to remove the `I` that was used for integral. Then the problem went away when we stopped using `using namespace std::complex`

SparseGrids/tasgridWrapper.cpp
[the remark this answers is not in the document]
mkstoyanov: Because `printMatrix()` is human readable and the format should be
```
<number> + <number> i  <number> + <number> i ...
```
I must have missed to add the comment regarding `printMatrix()`

SparseGrids/tsgLinearSolvers.cpp
zbmorrow: That was from `<complex.h>`, which defined `I` as a macro (http://pubs.opengroup.org/onlinepubs/009695399/basedefs/complex.h.html). The problem stopped when we changed `<complex.h>` to `<complex>`.

mkstoyanov:
* close/reopen to retest

SparseGrids/tasgridExternalTests.cpp
```diff
@@ -1333,6 +1411,21 @@ bool ExternalTester::testAllDomain() const{
                      pass2 = false;
             }
         }
+    }{
+        const BaseFunction *f = &f21coscos;
+        double errs[5] = {5.E-2, 7.E-3, 9.E-4, 1.E-13};
```
mkstoyanov: * formatting, add spaces after `{` and before `}`

SparseGrids/tasgridWrapper.cpp
```diff
@@ -1026,13 +1059,21 @@ void TasgridWrapper::writeMatrix(const char *filename, int rows, int cols, const
...
     cout << rows << " " << cols << endl;
     cout.precision(17);
     cout << std::scientific;
     for(size_t i=0; i<((size_t) rows); i++){
-        for(size_t j=0; j<((size_t) cols); j++){
-            cout << setw(25) << mat[i* ((size_t) cols) + j] << " ";
+        if (isComplex){
+            for(size_t j=0; j<((size_t) cols); j++){
+                cout << setw(25) << mat[i* ((size_t) cols) + ((size_t) 2*j)];   // real part
+                cout << (mat[i* ((size_t) cols) + ((size_t) 2*j+1)] < 0 ? " -" : " +");
```
mkstoyanov: * use `< 0.0`, i.e., specify that you are comparing floats

SparseGrids/tasgridWrapper.cpp
```diff
@@ -1026,13 +1059,21 @@ void TasgridWrapper::writeMatrix(const char *filename, int rows, int cols, const
...
+                cout << setw(24) << abs(mat[i* ((size_t) cols) + ((size_t) 2*j+1)]) << " i   ";     // imag part
```
mkstoyanov: * this is a bug, `abs()` works only for integer types, long, int, etc.
* you must use `fabs()` for floating point numbers
* at the end, why 2 spaces as opposed to 1?
* why not have `i` attached to the number, e.g., `1.234E-5 + 1.234.E-5i`

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -1056,6 +1067,23 @@ double* OneDimensionalNodes::getRLejaShifted(int n){
...
 
+double* OneDimensionalNodes::getFourierNodes(int level) {
+    OneDimensionalMeta meta;
+    int n = meta.getNumPoints(level, rule_fourier);
+    double* x = new double[n];
+    x[0]=0.0;
+    if(level > 0) { x[1]=1.0/3.0; x[2]=2.0/3.0; }
+    int count=3;
+    for(int l=2; l<=level; l++) {
+        n = meta.getNumPoints(l,rule_fourier);
```
mkstoyanov: * formatting, add space after the comma: `(l, rule_fourier)`

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -167,6 +167,69 @@ void TasmanianTridiagonalSolver::decompose(int n, double d[], double e[], double
     }
 }
 
+void TasmanianFourierTransform::dft(const int rank, const int n[], const std::complex<double> data[], std::complex<double> out[], const int direction){
```
mkstoyanov: * rename the `dft` function to something like: `discrete_fourier_transform`

if you google `dft code` the top hits are related to Density Field Theory, which is a huge field. Nobody is really using the simple Fourier transfrom, and we will drop it soon.

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -167,6 +167,69 @@ void TasmanianTridiagonalSolver::decompose(int n, double d[], double e[], double
...
+    // Fourier transform code by ZBM
+    // DOES NOT NORMALIZE the outgoing Fourier coefficients
+
+    if (direction != 1 && direction != -1) std::cerr << "ERROR: direction must be -1 (forward) or 1 (backward); value is " << direction << std::endl;
```
mkstoyanov: * do not use simple integers as flags in C++, it is OK for C or Fortran, but C++ has more elegant ways, e.g., boolean or enumerate both of which will make the code more readable
* do not change this now, the whole function will be scrapped anyway

mkstoyanov:
* added Fourier grids that are based on complex exponential basis

PR #50, opened 2018-07-11
Python interface for Fourier

zbmorrow:
This branch implements a Python wrapper for the Fourier basis. I've tested this out in a Jupyter notebook.

### DIFF
--- a/Config/AltBuildSystems/TasmanianSG.py
+++ b/Config/AltBuildSystems/TasmanianSG.py
@@ -661,12 +661,6 @@ class TasmanianSparseGrid:
                 controls the density of the grid, i.e.,
                 the offset for the tensor selection, the meaning of
                 iDepth depends on sType
-                Example 1: sType == 'iptotal' will give a grid that
-                           interpolates exactly all polynomials of
-                           degree up to and including iDepth
-                Example 2: sType == 'qptotal' will give a grid that
-                           integrates exactly all polynomials of degree
-                           up to and including iDepth
 
         sType: string identifying the tensor selection strategy
               'level'     'curved'     'hyperbolic'     'tensor'
@@ -1609,7 +1603,7 @@ class TasmanianSparseGrid:
         if (iNumPoints == 0):
             return np.empty([0,iNumOuts], np.float64)
 
-        if not self.isFourier():
+        if (not self.isFourier()):
             aSurp = np.empty([iNumOuts * iNumPoints], np.float64)
             self.pLibTSG.tsgGetHierarchicalCoefficientsStatic(self.pGrid, np.ctypeslib.as_ctypes(aSurp))
         else:
@@ -1727,7 +1721,7 @@ class TasmanianSparseGrid:
             raise TasmanianInputError("llfCoefficients", "ERROR: leading dimension of llfCoefficients is {0:1d} but the number of current points is {1:1d}".format(llfCoefficients.shape[0], self.getNumNeeded()))
         if (llfCoefficients.shape[1] != self.getNumOutputs()):
             raise TasmanianInputError("llfCoefficients", "ERROR: second dimension of llfCoefficients is {0:1d} but the number of outputs is set to {1:1d}".format(llfCoefficients.shape[1], self.getNumOutputs()))
-        if (self.isFourier() and not np.iscomplexobj(llfCoefficients)):
+        if (self.isFourier() and (not np.iscomplexobj(llfCoefficients))):
             raise TasmanianInputError("llfCoefficients", "ERROR: using Fourier grid but llfCoefficients is not complex")
 
         iNumPoints = llfCoefficients.shape[0]
--- a/Config/AltBuildSystems/TasmanianSG.py
+++ b/Config/AltBuildSystems/TasmanianSG.py
@@ -82,7 +82,7 @@ class TasmanianSimpleSparseMatrix:
     def getDenseForm(self):
         if ((self.iNumRows == 0) or (self.iNumCols == 0)):
             return np.empty([0,0], np.float64)
-        aMat = np.zeros([self.iNumRows, self.iNumCols], np.float64)
+        aMat = np.zeros([self.iNumRows, self.iNumCols], np.float64 if not np.iscomplexobj(self.aVals) else np.complex128)
         for iI in range(self.iNumRows):
             iJ = self.aPntr[iI]
             while(iJ < self.aPntr[iI+1]):
@@ -209,6 +209,7 @@ class TasmanianSparseGrid:
         self.pLibTSG.tsgIsSequence.argtypes = [c_void_p]
         self.pLibTSG.tsgIsLocalPolynomial.argtypes = [c_void_p]
         self.pLibTSG.tsgIsWavelet.argtypes = [c_void_p]
+        self.pLibTSG.tsgIsFourier.argtypes = [c_void_p]
         self.pLibTSG.tsgSetDomainTransform.argtypes = [c_void_p, POINTER(c_double), POINTER(c_double)]
         self.pLibTSG.tsgIsSetDomainTransfrom.argtypes = [c_void_p]
         self.pLibTSG.tsgClearDomainTransform.argtypes = [c_void_p]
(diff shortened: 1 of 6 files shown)


InterfacePython/TasmanianSG.in.py
```diff
@@ -644,6 +645,74 @@ def makeWaveletGrid(self, iDimension, iOutputs, iDepth, iOrder=1, liLevelLimits=
...
+
+        liAnisotropicWeights: list or numpy.ndarray of weights
+                              length must be iDimension or 2*iDimension
+                              the first iDimension weights
+                                                       must be positive
+                              see the manual for details
+
+        '''
+        if (iDimension <= 0):
+            raise TasmanianInputError("iDimension", "ERROR: dimension should be a positive integer")
```
mkstoyanov: * the exceptions must be tested
* inside `testsTSG.in.py` look for function `def testBasicException()`. You want to essentially repeat the first group of tests that relates to `makeSequenceGrid()`, but make it for `def makeFourierGrid()`

InterfacePython/TasmanianSG.in.py
```diff
@@ -644,6 +645,74 @@ def makeWaveletGrid(self, iDimension, iOutputs, iDepth, iOrder=1, liLevelLimits=
...
+            if (len(liLevelLimits) != iDimension):
+                raise TasmanianInputError("liLevelLimits", "ERROR: invalid number of level limits, must be equal to iDimension")
+            pLevelLimits = (c_int*iDimension)()
+            for iI in range(iDimension):
+                pLevelLimits[iI] = liLevelLimits[iI]
+
+        if (sys.version_info.major == 3):
+            sType = bytes(sType, encoding='utf8')
+
+        self.pLibTSG.tsgMakeFourierGrid(self.pGrid, iDimension, iOutputs, iDepth, c_char_p(sType), pAnisoWeights, pLevelLimits)
```
mkstoyanov: * you should test one call to the grid to see if it returns the correct set of points. Use pen and paper to write down the points (in Tasmanian order) for a small grid (2D with 10 - 20 points). Then, inside `testsTSG.in.py` add a test inside `def testFullCoverageA()` that makes a grid, reads the points and compares.

SparseGrids/TasmanianSparseGrid.cpp
```diff
@@ -1585,6 +1585,14 @@ void tsgMakeLocalPolynomialGrid(void *grid, int dimensions, int outputs, int dep
 void tsgMakeWaveletGrid(void *grid, int dimensions, int outputs, int depth, int order, const int *limit_levels){
     ((TasmanianSparseGrid*) grid)->makeWaveletGrid(dimensions, outputs, depth, order, limit_levels);
 }
+void tsgMakeFourierGrid(void *grid, int dimensions, int outputs, int depth, const char *sType, const int *anisotropic_weights, const int *limit_levels){
```
mkstoyanov: * the C interface is good, but must test the grid I/O as well
* inside `testsTSG.in.py`, look for `def testBasicIO()`, between the tests for wavelets and conformal transforms add a few tests regarding the read/write in different formats

zbmorrow:
I wasn't 100% sure if modifying `GridFourier::integrate()` was appropriate for this pull request. However, if I left that alone, then I would have had to load the function values into both grids in the Python tests of `getHierarchicalCoefficients`, which would give an incorrect test result if `getHierarchicalCoefficients` does nothing. I guess I could have used `np.testing.assert_equal`, though...

SparseGrids/tsgGridFourier.cpp
```diff
@@ -735,17 +735,29 @@ void GridFourier::evaluateBatchGPUmagma(const double x[], int num_x, double y[],
 }
 
 void GridFourier::integrate(double q[], double *conformal_correction) const{
```
mkstoyanov: * make a separate pull-request, this is a bug that slipped the C++ tests
* check `GridSequence::integrate()` for reference (switch the if-statement)
* there is no need of OpenMP here, `integrate` is something done once and the speedup would be marginal anyway

zbmorrow:
Rebased and verified that no modified files have trailing whitespaces

Config/AltBuildSystems/TasmanianSG.py
```diff
@@ -644,6 +646,74 @@ def makeWaveletGrid(self, iDimension, iOutputs, iDepth, iOrder=1, liLevelLimits=
...
+        iOutputs: int (non-negative)
+              the number of outputs
+
+        iDepth: int (non-negative)
+                controls the density of the grid, i.e.,
+                the offset for the tensor selection, the meaning of
+                iDepth depends on sType
+                Example 1: sType == 'iptotal' will give a grid that
+                           interpolates exactly all polynomials of
+                           degree up to and including iDepth
```
mkstoyanov: * this relates to `OneDimensionalMeta::getIExact()` and `OneDimensionalMeta::getQExact()`, which currently default Fourier grids to just `level`
* remove the two examples from the comment, we will make examples later once we figure out the frequencies
* you have to change the main file and then regenerate the windows/GNU Make files

Config/AltBuildSystems/TasmanianSG.py
```diff
@@ -1634,9 +1727,21 @@ def setHierarchicalCoefficients(self, llfCoefficients):
             raise TasmanianInputError("llfCoefficients", "ERROR: leading dimension of llfCoefficients is {0:1d} but the number of current points is {1:1d}".format(llfCoefficients.shape[0], self.getNumNeeded()))
         if (llfCoefficients.shape[1] != self.getNumOutputs()):
             raise TasmanianInputError("llfCoefficients", "ERROR: second dimension of llfCoefficients is {0:1d} but the number of outputs is set to {1:1d}".format(llfCoefficients.shape[1], self.getNumOutputs()))
+        if (self.isFourier() and not np.iscomplexobj(llfCoefficients)):
```
mkstoyanov: * be explicit about brackets in if-statments
```
if (self.isFourier() and (not np.iscomplexobj(llfCoefficients))):
```

Config/AltBuildSystems/TasmanianSG.windows.py
```diff
@@ -644,6 +646,74 @@ def makeWaveletGrid(self, iDimension, iOutputs, iDepth, iOrder=1, liLevelLimits=
...
+              the number of inputs
+
+        iOutputs: int (non-negative)
+              the number of outputs
+
+        iDepth: int (non-negative)
+                controls the density of the grid, i.e.,
+                the offset for the tensor selection, the meaning of
+                iDepth depends on sType
+                Example 1: sType == 'iptotal' will give a grid that
```
mkstoyanov: * same as before, remove the examples

Config/AltBuildSystems/TasmanianSG.py
```diff
@@ -1134,6 +1204,12 @@ def isWavelet(self):
         '''
         return (self.pLibTSG.tsgIsWavelet(self.pGrid) != 0)
 
+    def isFourier(self):
+        '''
+        returns True if using a Fourier grid
+        '''
+        return (self.pLibTSG.tsgIsFourier(self.pGrid) != 0)
```
mkstoyanov: * nothing wrong, just giving a comment
* in C programming language, when integers are used as booleans, e.g.,
```
int a;
...
if (a) ...
```
then `0` is interpreted as `false`, while any non-zero value is `1`. It is better to make the check explicit, but always stick to the convention:
```
if (a != 0)
```
every time integers are used as flags for booleans.

Config/AltBuildSystems/TasmanianSG.py
```diff
@@ -1532,8 +1608,15 @@ def getHierarchicalCoefficients(self):
         iNumPoints = self.getNumLoaded()
         if (iNumPoints == 0):
             return np.empty([0,iNumOuts], np.float64)
-        aSurp = np.empty([iNumOuts * iNumPoints], np.float64)
-        self.pLibTSG.tsgGetHierarchicalCoefficientsStatic(self.pGrid, np.ctypeslib.as_ctypes(aSurp))
+
+        if not self.isFourier():
```
mkstoyanov: * add brackets for consistency
```
if (not self.isFourier()):
```

PR #51, opened 2018-07-11
MATLAB interface for Fourier

zbmorrow:
This branch adds a MATLAB wrapper for the Fourier basis. Changes:
* `tsgReadMatrix.m` now has a boolean flag `bIsComplex`; `tsgWriteMatrix.m` automatically detects whether the matrix is complex
* added `tsgMakeFourier.m`
* fixed a bug that was causing all-zero quadrature weights for the Fourier rule when using `./tasgrid -makequadrature` (source was in `tsgCoreOneDimensional.cpp`)
* added extra case to `printGridStats` in `TasmanianSparseGrid.cpp` so that the MATLAB interface can detect a Fourier rule
* small formatting changes

### DIFF
--- a/InterfaceMATLAB/tsgCoreTests.m
+++ b/InterfaceMATLAB/tsgCoreTests.m
@@ -464,21 +464,21 @@ if ((abs(norm(sum(w)) - 58.0) > 1.E-11) || (abs(max(p(:, 1)) - 13.0/3.0) > 1.E-1
 end
 
 % correctness of 1-D
-[lGrid, pf] = tsgMakeFourier('_tsgcoretests_lgrid', 1, 1, 'level', 2);
-bmark = [0.0; 1.0/3.0; 2.0/3.0; 1.0/9.0; 2.0/9.0; 4.0/9.0; 5.0/9.0; 7.0/9.0; 8.0/9.0];
-if (norm(pf - bmark) > 1.E-11)
+[lGrid, p] = tsgMakeFourier('_tsgcoretests_lgrid', 1, 1, 'level', 2);
+tp = [0.0; 1.0/3.0; 2.0/3.0; 1.0/9.0; 2.0/9.0; 4.0/9.0; 5.0/9.0; 7.0/9.0; 8.0/9.0];
+if (norm(p - tp) > 1.E-11)
     error('Mismatch in tsgMakeFourier: points');
 end
 
 % level limits
-[lGrid, p] = tsgMakeFourier('_tsgcoretests_lgrid', 3, 1, 'level', 2, [], [], [], [], [0, 1, 2]);
-if (min(abs(p(:,1))) > 1.E-8)
+[lGrid, p] = tsgMakeFourier('_tsgcoretests_lgrid', 3, 1, 'level', 3, [], [], [], [], [0, 1, 2]);
+if (max(abs(p(:,1))) > 1.E-8)
     error('Mismatch in tsgMakeFourier: level limit, dim 1');
 end
-if (min(abs(p(:,2) - 2.0/3.0)) > 1.E-8)
+if (min(abs(p(:,2) - 1.0/9.0)) < 1.E-8)
     error('Mismatch in tsgMakeFourier: level limit, dim 2');
 end
-if (min(abs(p(:,3) - 8.0/9.0)) > 1.E-8)
+if (min(abs(p(:,3) - 1.0/27.0)) < 1.E-8)
     error('Mismatch in tsgMakeFourier: level limit, dim 3');
 end
 tsgDeleteGrid(lGrid);
--- a/InterfaceMATLAB/tsgCoreTests.m
+++ b/InterfaceMATLAB/tsgCoreTests.m
@@ -445,6 +445,44 @@ if (min(abs(p(:,3) - 0.125)) < 1.E-8)
 end
 tsgDeleteGrid(lGrid);
 
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%%%                     tsgMakeFourier()                             %%%
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+% test transform
+[lGrid, p] = tsgMakeFourier('_tsgcoretests_lgrid', 3, 1, 'level', 1, [3.0 5.0; -7.0 -6.0; -12.0 17.0]);
+if ((abs(max(p(:, 1)) - 13.0/3.0) > 1.E-11) || (abs(min(p(:, 1)) - 3.0) > 1.E-11) ...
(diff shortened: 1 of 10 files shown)


dalg24:
ok to test

(This should not be necessary next time.  I added @zbmorrow to the white-list per @mkstoyanov request.)

InterfaceMATLAB/tsgGetHCoefficients.m
```diff
@@ -34,7 +34,8 @@
         fprintf(1,['Warning: Command had non-empty output:\n']);
         disp(cmdout);
     end
-    [coefficients] = tsgReadMatrix(sFileO);
+    
+    [coefficients] = tsgReadMatrix(sFileO, strcmp(lGrid.sType, 'fourier'));  % read file as a complex file if appropriate
```
mkstoyanov: * Is this the only place where complex numbers are used? If so, you should not change `tsgReadMatrix()` but make the coefficient conversion here.
* MATLAB by default uses `i` for complex number if you have not used it for loop index or something else (if I remember right)
* Also, you are using for loops inside MATLAB which is horribly inefficient, you must always vectorize MATLAB code:
```
if (strcmp(lGrid.sType, 'fourier'))
    coefficients = coefficients(:, 1:2:end) + i * coefficients(:, 2:2:end);
end
```

InterfaceMATLAB/tsgMakeFourier.m
```diff
@@ -0,0 +1,221 @@
+function [lGrid, points] = tsgMakeFourier(sGridName, iDim, iOut, sType, iDepth, mTransformAB, vAnisotropy, sConformalMap, vConfromalWeights, vLimitLevels)
```
mkstoyanov: * this needs testing is `tsgCoreTests()`, see how other grids are tested and adjust the tests. Keep things simple.
* make sure to add a test regarding the complex numbers and the matrix.

InterfaceMATLAB/tsgWriteMatrix.m
```diff
@@ -13,28 +13,56 @@ function tsgWriteMatrix(filename, mat)
...
+% 9 10 11 12
+%
+
+bIsComplex = ~isreal(mat);
 
 if (prod(size(mat)) < 1000) % small matrix, use ascii format
 
     fid = fopen(filename, 'w');
 
-    fprintf(fid, '%d  %d\n', size(mat, 1), size(mat, 2)); % load the number of points
```
mkstoyanov: * same as `tsgReadMatrix()`, this is only ever called on `setHierarchicalCoefficients()`, in that function you should first make sure that complex numbers are used only when the grid is `fourier` and Global or Sequence. You should also convert the matrix to reals and pass only a single double matrix to `tsgWriteMatrix()`
```
real_matrix = [complex_matrix, complex_matrix]; % defines the dimension
real_matrix(:, 1:2:end) = real(complex_matrix);
real_matrix(:, 2:2:end) = imag(complex_matrix);
```

InterfaceMATLAB/tsgCoreTests.m
```diff
@@ -445,6 +445,44 @@ function tsgCoreTests()
...
+[w, p] = tsgGetQuadrature(lGrid);
+if ((abs(norm(sum(w)) - 58.0) > 1.E-11) || (abs(max(p(:, 1)) - 13.0/3.0) > 1.E-11) ...
+    || (abs(min(p(:, 1)) - 3.0) > 1.E-11) || (abs(max(p(:, 2)) + 19.0/3.0) > 1.E-11) ...
+    || (abs(min(p(:, 2)) + 7.0) > 1.E-11) || (abs(max(p(:, 3)) - 22.0/3.0) > 1.E-11) ...
+    || (abs(min(p(:, 3)) + 12.0) > 1.E-11))
+    error('Mismatch in tsgMakeFourier: getQuadrature');
+end
+
+% correctness of 1-D
+[lGrid, pf] = tsgMakeFourier('_tsgcoretests_lgrid', 1, 1, 'level', 2);
```
mkstoyanov: * why use `pf` as opposed to just `p`
* also, why use `bmark` as opposed to `tp`, I use `tp` in tests to indicate `true points`
* consistent variable names make the code easier to read

InterfaceMATLAB/tsgCoreTests.m
```diff
@@ -445,6 +445,44 @@ function tsgCoreTests()
...
+bmark = [0.0; 1.0/3.0; 2.0/3.0; 1.0/9.0; 2.0/9.0; 4.0/9.0; 5.0/9.0; 7.0/9.0; 8.0/9.0];
+if (norm(pf - bmark) > 1.E-11)
+    error('Mismatch in tsgMakeFourier: points');
+end
+
+% level limits
+[lGrid, p] = tsgMakeFourier('_tsgcoretests_lgrid', 3, 1, 'level', 2, [], [], [], [], [0, 1, 2]);
+if (min(abs(p(:,1))) > 1.E-8)
```
mkstoyanov: * if you are looking if there is a point other than `0`, shouldn't this be `max(abs...`

InterfaceMATLAB/tsgCoreTests.m
```diff
@@ -445,6 +445,44 @@ function tsgCoreTests()
...
+    error('Mismatch in tsgMakeFourier: level limit, dim 1');
+end
+if (min(abs(p(:,2) - 2.0/3.0)) > 1.E-8)
```
mkstoyanov: * this will be false if `2.0/3.0` is missing from the grid, the point of level limits is to ensure that there are no points from an upper level, i.e., you should look if `1.0/9.0` is included by mistake.

InterfaceMATLAB/tsgEvaluateHierarchy.m
```diff
@@ -57,6 +57,9 @@
     end
     if (strcmp(lGrid.sType, 'global') || strcmp(lGrid.sType, 'sequence'))
         [vals] = tsgReadMatrix(sFileO);
+    elseif (strcmp(lGrid.sType, 'fourier'))
+        [tmp] = tsgReadMatrix(sFileO);
+        vals = tmp(:,1:2:end) + sqrt(-1)*tmp(:,2:2:end);
```
mkstoyanov: * why not use MATLAB's own `i`, technically `sqrt(-1)` could be `+i` or `-i`

InterfaceMATLAB/tsgGetHCoefficients.m
zbmorrow: * Were there cases after commit `4a714cd` where I didn't vectorize? I thought I had changed all my old for-loops, but I may have missed one.
* You are right about `i`. I used `sqrt(-1)` just to be extra safe...

InterfaceMATLAB/tsgEvaluateHierarchy.m
mkstoyanov: @zbmorrow: let's be consistent with notation, we are using `i` everywhere to write to `cout` and in other functions. Change the loop counter from `i` to `r`.

zbmorrow:
I have replicated all the modified C++ files over to the `fourier_fixes` pull request. I left the modifications to `TasmanianSparseGrid` and `tsgCoreOneDimensional` here so that the MATLAB tests would run.

zbmorrow:
Rebased after merge of pull request #56 into `master`

PR #58, opened 2018-07-14
Cmake creates matlab folder

mkstoyanov:
* have cmake create the MATLAB work folder (if missing)
* the install script uses only the work folder, no need for an On/Off variable
* the install script still creates the MATLAB work and converts relative paths (relative to the source folder) to absolute path


mkstoyanov:
* have cmake create the MATLAB work folder (if missing)
* the install script uses only the work folder, no need for an On/Off variable
* the install script still creates the MATLAB work and converts relative paths (relative to the source folder) to absolute path

PR #61, opened 2018-07-18
FFT for N=3^k points

zbmorrow:
This branch implements an O(N log N) recursive FFT for N_d=3^{k_d} points in each direction. I've compared the outputs of this FFT for 2 and 3 dimensions against MATLAB and the FFTW library.


mkstoyanov:
* this needs a discussion

PR #75, opened 2018-07-25
Fortran interface and tests

vreshniak:
Most of the Fortran tests are translated from the corresponding Matlab tests.

Important changes to the interface include:
  * Now Fortran interface does not require for the explicit calls to tsgInitialize() and tsgFinalize()
  * the parameter "output" in functions tsgSetAnisotropicCoefficients(), tsgEstimateAnisotropicCoefficients(), tsgSetGlobalSurplusRefinement(), tsgSetLocalSurplusRefinement() now assumes values >= 1


mkstoyanov:
OK to test.

InterfaceFortran/fortester.f90
```diff
@@ -19,192 +19,1518 @@
...
+  double precision :: res
+
+  res = sqrt(sum(abs(x)**2))
+end function 
+
+function norm1d(x) result(res)
+  double precision :: x(:)
+  double precision :: res
+
+  res = sqrt(sum(abs(x)**2))
```
mkstoyanov: * why is the `abs()` needed? Real numbers can just use `x**2`

InterfaceFortran/TasmanianSG.f90
```diff
@@ -154,501 +166,646 @@ Module TasmanianSG
...
-END SUBROUTINE tsgSetConformalTransformASIN
-!=======================================================================
-FUNCTION tsgIsSetConformalTransformASIN(gridID) result(isset)
-  INTEGER :: gridID, res
-  LOGICAL :: isset
-  CALL tsgica(gridID, res)
-  IF (res .EQ. 0)then
+  endif
+  if(present(levelLimits))then
+    call tsgshr(gridID, tolerance, rtype, theout, levelLimits)
```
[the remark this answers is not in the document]
mkstoyanov: * take all pointers that have optional values, e.g., can be `null` and add a flag in the C wrappers
* if the flag is true, use the value provided by the Fortran module, if false pass to C++ the `null` pointer
* this will cover *make* functions and *refine* functions, and refine
* double-check to see if there are other functions that take default values

InterfaceFortran/fortester.f90
```diff
@@ -19,192 +19,1518 @@
...
+! initialize random number generator
+call srand(seed)
+
+
+
+!=======================================================================
+!       tsgInitialize()
+!       tsgFinalize()
+!=======================================================================
+i_a = 8
```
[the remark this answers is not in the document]
mkstoyanov: * move this to the `tsgTestInternals()`

InterfaceFortran/fortester.f90
mkstoyanov: * remove the `abs()` things cannot be complex here

CHANGELOG.md
```diff
@@ -1,23 +1,43 @@
 Changelog for version 6.0
 --------------
 
+* added new grids and rules
```
mkstoyanov: * @vreshniak: at some point you make `pull` instead of `rebase`, which is now doubling up on many of changes that are already in the main (e.g., the changelog or fixed memory leaks or the use of `std::vector`).
* this is also going to make the review very hard, I'll give you comments on the Fortran module and wrapper, but the commit history will need a lot of work before it can be accepted
* let me know if you need help

vreshniak: I was going to ask you about this.
I wasn't able to push changes after rebase without pulling first.
I guess I can force push to resolve this issue.
Can we talk about this today?

mkstoyanov: I had a long meeting this morning, I will be in my office for the rest of the day.

InterfaceFortran/TasmanianSG.f90
```diff
@@ -154,501 +169,732 @@ Module TasmanianSG
...
+  integer :: rows, cols, length
+  double precision, pointer :: matrix(:,:), vector(:)
+  character, pointer :: string(:)
+contains
+!=======================================================================
+function tsgIsGlobal(id) result(res)
+  integer :: id, flag
+  logical :: res
+  call tsgisg(id, flag)
+  if (flag .eq. 1) then
```
mkstoyanov: * `tsgisg()` is a C function returning `flag` which is in fact an integer, basically an integer is used for a boolean/logical
* in C programming standard, integer `0` is considered `false` while `not-0` is considered `true`
* when using an integer as a C boolean/logical, you should never have statement such as `(flag .eq. 1)` as flag of 2, 3, and so on, is also technically `true`
* you wrote the `tsgisg()` and you know what it returns, but conform to the C standard, i.e., use
```
if (flag .ne. 0)
   res = .true.
...
```
or
```
if (flag .eq. 0)
  res = .false.
...
```

mkstoyanov: * fix the rest of the functions too

InterfaceFortran/TasmanianSG.f90
```diff
@@ -154,501 +166,646 @@ Module TasmanianSG
...
-  CALL tsgglp(gridID, p)
-END FUNCTION tsgGetLoadedPoints
+  allocate(p(rows,cols))
+  call tsgglp(gridID, p)
+end function tsgGetLoadedPoints
 !=======================================================================
-FUNCTION tsgGetNeededPoints(gridID) result(p)
-  DOUBLE PRECISION, pointer :: p(:,:)
-  INTEGER :: gridID, rows, cols
+function tsgGetNeededPoints(gridID) result(p)
```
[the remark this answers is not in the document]
mkstoyanov: * cannot use allocatable, got it!

InterfaceFortran/TasmanianSG.f90
```diff
@@ -154,501 +166,646 @@ Module TasmanianSG
...
+  integer :: gridID, numX
+  double precision :: x(:,:), y(:,:)
+  call tsgevb(gridID, x, numX, y)
+end subroutine tsgEvaluateBatch
+!=======================================================================
+subroutine tsgEvaluateHierarchicalFunctions(gridID, x, numX, y)
+  integer :: gridID, numX
+  double precision :: x(:,:), y(:,:)
+  double precision :: y_c_style(size(y,2),size(y,1))
+  call tsgehf(gridID, x, numX, y_c_style)
```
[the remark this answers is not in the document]
mkstoyanov: * if the grid is not Fourier, then call the C function, otherwise print an error message 
```
ERROR: called tsgEvaluateHierarchicalFunctions() on a Fourier grid, use tsgEvaluateComplexHierarchicalFunctions() instead
```
We will deal with rigorous error-checking and message passing later.

InterfaceFortran/TasmanianSG.f90
```diff
@@ -154,501 +166,646 @@ Module TasmanianSG
...
+  y = transpose(y_c_style)
+end subroutine tsgEvaluateHierarchicalFunctions
+!=======================================================================
+subroutine tsgEvaluateComplexHierarchicalFunctions(gridID, x, numX, y)
```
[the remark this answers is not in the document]
mkstoyanov: * same as above

InterfaceFortran/TasmanianSG.f90
```diff
@@ -154,501 +169,732 @@ Module TasmanianSG
...
+  allocate(truncate(N))
+  call tsgcre(gridID, truncate)
+end function tsgGetConformalTransformASIN
+!=======================================================================
+subroutine tsgPrintStats(gridID)
+  integer :: gridID
+  call tsgpri(gridID)
+end subroutine tsgPrintStats
+!=======================================================================
+function tsgTestInternals(verbose) result(res)
```
mkstoyanov: * If I read this correctly, the only way this test can pass is if it gets called before any `tsgNewGridID()`, is that correct?
* You can make the first command of the test to be `tsgClearAll()` and add comment that this will delete all grids.

InterfaceFortran/fortester.f90
```diff
@@ -27,184 +27,1483 @@
...
+double complex,   allocatable :: dcmplx_2d_a(:,:), dcmplx_2d_b(:,:), dcmplx_2d_c(:,:), dcmplx_2d_d(:,:)
+
+integer,          pointer :: int_pnt_1d_a(:),    int_pnt_1d_b(:),    int_pnt_1d_c(:)
+double precision, pointer :: double_pnt_1d_a(:), double_pnt_1d_b(:), double_pnt_1d_c(:)
+double complex,   pointer :: dcmplx_pnt_1d_a(:), dcmplx_pnt_1d_b(:), dcmplx_pnt_1d_c(:)
+
 
 ! must call tsgInitialize() once per program
-  CALL tsgInitialize()
+! call tsgInitialize()
```
mkstoyanov: * delete this line and the comment above

InterfaceFortran/fortester.f90
```diff
@@ -27,184 +27,1483 @@
...
+! call tsgMakeLocalPolynomialGrid(gridID_II, 2, 1, 5, 1, tsg_semi_localp)
+! points  => tsgGetPoints(gridID);  i_b = tsgGetNumPoints(gridID)
+! rnd     => rand(2,i_a)
+! pointsb =  -1.d0 + 2.d0 * rnd
+! allocate( double_2d_a(1,i_b), double_1d_b(i_b), double_2d_d(i_b,i_b) )
+! double_2d_a(1,:) = exp( -points(1,:)**2 - 2.d0 * points(2,:)**2 )
+! call tsgLoadNeededPoints(gridID, double_2d_a)
+! call tsgEvaluateHierarchicalFunctions(gridID,points,i_b,double_2d_d)
+! double_1d_b = tsgSolveLeastSquares(double_2d_d,double_2d_a,0.d0)
+! call tsgSetHierarchicalCoefficients(gridID_II,double_1d_b)
```
mkstoyanov: * So `tsgSetHierarchicalCoefficients()` is not tested right now? Here is the test:
```
call tsgMakeLocalPolynomialGrid(gridID,    2, 1, 5, 1, tsg_semi_localp)
call tsgMakeLocalPolynomialGrid(gridID_II, 2, 1, 5, 1, tsg_semi_localp)
...
call tsgLoadNeededPoints(gridID, double_2d_a)
double_1d_b = tsgGetHierarchicalCoefficients(gridID)
call tsgSetHierarchicalCoefficients(gridID_II,double_1d_b)
call tsgEvaluateBatch(gridID,pointsb,i_a,double_2d_b)
call tsgEvaluateBatch(gridID_II,pointsb,i_a,double_2d_c)
if ( norm2d(double_2d_b - double_2d_c) > 1.d-11 ) then
  ...
```
Basically, instead of using leas-squares, use another grid to compute the coefficients and see if you can properly copy-paste them from one grid to the next.

* remove the commented code

InterfaceFortran/tsgC2Fortran.cpp
```diff
@@ -104,30 +116,73 @@ void tsgwri_(int *id, const char *filename, int *binary){ _tsg_grid_list[*id]->w
...
 
 // create
-void tsgmg_(int *id, int *dimensions, int *outputs, int *depth, int *type, int *rule, const int *anisotropic_weights, double *alpha, double *beta, const int *llimits){
+void tsgmg_(int *id, int *dimensions, int *outputs, int *depth, int *type, int *rule, int *opt_flags,
+            const int *aniso_weights, double *alpha, double *beta, const char *customRuleFilename, const int *llimits){
+    double al, be;
+    const int  *aw, *ll;
+    const char *cfn;
+
+    opt_flags[0] == 1 ? aw  = aniso_weights : aw = 0;
```
mkstoyanov: * same as above, you should compare to `0` and not `1`, e.g., `opt_flags[0] != 0`

InterfaceFortran/tsgC2Fortran.cpp
```diff
@@ -104,30 +116,73 @@ void tsgwri_(int *id, const char *filename, int *binary){ _tsg_grid_list[*id]->w
...
+    opt_flags[1] == 1 ? al  = *alpha : al = 0;
```
mkstoyanov: why not:
```
al = (opt_falgs[1] != 0) *alpha : 0.0;
```
* this makes it explicit that you are assigning something to `al`
* also, to avoid confusion, I always use `0.0` for floats

InterfaceFortran/Examples/example_sparse_grids.f90
```diff
@@ -92,7 +92,7 @@ PROGRAM TasmanianSGExample
 ! using classical Smolyak grid with Clenshaw-Curtis points and weights
 
 ! must call tsgInitialize() once per program
-  CALL tsgInitialize()
+!  CALL tsgInitialize()
```
mkstoyanov: * delete this and other similar statements

InterfaceFortran/TasmanianSG.f90
mkstoyanov: * Yes, I want to be compliant with the C interface. I know this is not a conflict with your code, but it can confuse someone else. We should always comply with the standards of various languages, whenever we can and whenever we know what the actual standard is. We know this one standard.
* In most cases you can just change `.eq.` with `.ne.` or `==` to `!=`

InterfaceFortran/TasmanianSG.f90
vreshniak: * This is true
* Instead I can call tsggag(num_ag) first and use it as a base value

mkstoyanov: * Fix it either way, use a base value or call `tsgClearAll()` with an added comment

InterfaceFortran/tsgC2Fortran.cpp
vreshniak: This code also does not rely on C interpretation of true/false.
Do you want me to make everything be consistent with the C standard?

InterfaceFortran/tsgC2Fortran.cpp
vreshniak: You are right, this is better style.

InterfaceFortran/tsgC2Fortran.cpp
mkstoyanov: * Yes, it is not in many places and it is OK if you skip a spot or two, but we should be compliant everywhere.

InterfaceFortran/TasmanianSG.f90
vreshniak: However, we test tsgClearAll() at the end anyway

mkstoyanov: You decide, `tsgClearAll()` should be called at least once, and `tsgTestInternals()` should either not-crash or add comment about why it is crashing.

vreshniak: Also, current implementation of tsgEvaluateHierarchicalFunctions() allows one to call it on Fourier grid. 
The result is real array with twice as many elements.
Do you want me to change this?

mkstoyanov: Yes, if we have complex evaluate, there is no need to keep the other one.

InterfaceFortran/fortester.f90
```diff
@@ -1397,7 +1396,7 @@ program FORTESTER
 
 contains
 
-function rand(n,m) result(res)
+function random(n,m) result(res)
```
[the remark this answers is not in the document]
vreshniak: There is intrinsic rand() function.
I changed it to avoid confusion

InterfaceFortran/tsgC2Fortran.cpp
```diff
@@ -104,30 +116,73 @@ void tsgwri_(int *id, const char *filename, int *binary){ _tsg_grid_list[*id]->w
...
-void tsgmw_(int *id, int *dimensions, int *outputs, int *depth, int *order){
-    _tsg_grid_list[*id]->makeWaveletGrid(*dimensions, *outputs, *depth, *order);
+void tsgmw_(int *id, int *dimensions, int *outputs, int *depth, int *opt_flags, int *order, const int *llimits){
+    int ord;
+    const int *ll;
+
+    ord = (opt_flags[0] != 0) ? *order  : 1;
+    ll  = (opt_flags[1] != 0) ? llimits : 0;
+
+    _tsg_grid_list[*id]->makeWaveletGrid(*dimensions, *outputs, *depth, *order, llimits);
```
mkstoyanov: * bug, this should be:
```
_tsg_grid_list[*id]->makeWaveletGrid(*dimensions, *outputs, *depth, ord, ll);
```

InterfaceFortran/fortester.f90
```diff
@@ -27,184 +27,1471 @@
...
+     check_points( points(2,:), (/0.d0,-1.d0,1.d0/), (/1.d0/sqrt(3.d0)/) ) .or. &
+     check_points( points(3,:), (/0.d0,1.d0/), (/-1.d0,1.d0/sqrt(3.d0)/) ) ) then
+  write(*,*) "Mismatch in tsgSetAnisotropicRefinement: limits in make"
+  stop 1
+endif
+allocate( double_2d_a(1,i_b) )
+double_2d_a(1,:) = exp( -points(1,:)**2 - points(2,:)**2 )
+call tsgLoadNeededPoints(gridID, double_2d_a)
+call tsgSetAnisotropicRefinement( gridID, tsg_iptotal, 30, 1 )
+pointsb => tsgGetNeededPoints(gridID)
```
mkstoyanov: * must delete `pointsb` after the checks and before you call `pointsb =>` again in line 1230.

InterfaceFortran/fortester.f90
```diff
@@ -27,184 +27,1471 @@
...
+
+points => tsgGetNeededPoints(gridID);  i_b = tsgGetNumNeeded(gridID)
+allocate( double_2d_a(1,i_b) )
+double_2d_a(1,:) = exp( -points(1,:)**2 - points(2,:)**2 )
+call tsgLoadNeededPoints(gridID, double_2d_a)
+call tsgMergeRefinement(gridID_II)
+deallocate(points, double_2d_a)
+
+i_a = 20
+allocate( pointsb(2,i_a) );
```
mkstoyanov: * here you're allocating `pointsb` only to then have it immediately reassigned in line 1349. Get rid of this line.

InterfaceFortran/fortester.f90
```diff
@@ -27,184 +27,1471 @@
...
+
+points  => tsgGetPoints(gridID_II);     i_b = tsgGetNumPoints(gridID_II)
+allocate( double_2d_a(1,i_b) )
+double_2d_a(1,:) = exp( -points(1,:)**2 - points(2,:)**2 )
+call tsgLoadNeededPoints(gridID_II, double_2d_a)
+deallocate(points, double_2d_a)
+
+i_a = 30
+allocate( points(2,i_a), double_2d_a(1,i_a), double_2d_b(1,i_a) )
+rnd     => random(2,i_a)
```
mkstoyanov: deallocate `rnd` after you assign `points=`

InterfaceFortran/fortester.f90
```diff
@@ -27,184 +27,1471 @@
...
+
+  int  = sum( weights  * ( 1.d0/((1.d0+5.d0*points(1,:)**2)*(1.d0+5.d0*points(2,:)**2)) ) )
+  int2 = sum( weightsb * ( 1.d0/((1.d0+5.d0*pointsb(1,:)**2)*(1.d0+5.d0*pointsb(2,:)**2)) ) )
+
+  if ( abs(int-1.028825601981092d0**2) < abs(int2 - 1.028825601981092d0**2) ) then
+    write(*,*) "Mismatch in tsgMakeWaveletGrid: conformal map"
+    stop 1
+  endif
+enddo
+deallocate(points, pointsb, weights, weightsb)
```
mkstoyanov: the deallocate should be inside of the loop

InterfaceFortran/fortester.f90
```diff
@@ -27,184 +27,1471 @@
...
+i_b = 13
+allocate(double_2d_a(1,i_a), double_2d_b(1,i_b), double_2d_c(i_b,2*i_a), &
+         double_1d_a(i_b),   double_1d_b(i_b),   double_1d_c(i_b), pointsb(2,i_b), double_1d_d(2*i_a))
+double_2d_a(1,:) = exp( -points(1,:)**2 - 2.d0 * points(2,:)**2 )
+call tsgLoadNeededPoints(gridID, double_2d_a)
+rnd     => random(2,i_b)
+pointsb = -1.d0 + 2.d0 * rnd
+call tsgEvaluateBatch(gridID,pointsb,i_b,double_2d_b)
+
+allocate(dcmplx_1d_a(i_a),dcmplx_2d_c(i_a,i_b))
```
mkstoyanov: deallocate these at some point, after you are done with the two matrices

InterfaceFortran/fortester.f90
```diff
@@ -27,184 +27,1471 @@
...
+enddo
+deallocate(double_2d_a, double_2d_b, double_2d_c, double_1d_a, double_1d_b, pointsb, points)
+
+
+call tsgMakeGlobalGrid(gridID, 2, 1, 22, tsg_iptotal, tsg_chebyshev)
+points => tsgGetPoints(gridID)
+allocate(double_2d_a(1,tsgGetNumPoints(gridID)))
+double_2d_a(1,:) = exp( -points(1,:)**2 - points(2,:)**2 )
+call tsgLoadNeededPoints(gridID, double_2d_a)
+allocate(pointsb(2,1000),double_2d_b(1,1000),double_2d_c(1,1000))
```
mkstoyanov: you are setting `pointsb` here for evaluation, but you are not initializing with random values

PR #83, opened 2018-07-28
speedup tests on multi-gpu systems

mkstoyanov:
when dealing with multiple gpus, the reference cpu solution is computed only once (switched order of loops)


mkstoyanov:
* troubles with the CI system, no need to stress the server, will reopen once resolved (and debugged)

PR #95, opened 2018-07-31
cmake now keeps track of changes in the matlab source files

mkstoyanov:


InterfaceMATLAB/CMakeLists.txt
```diff
@@ -3,45 +3,50 @@
...
+                                  tsgRefineAnisotropic.m
+                                  tsgRefineSurplus.m
+                                  tsgReloadGrid.m
+                                  tsgSummary.m
+                                  tsgCoreTests.m
+                                  tsgWriteCustomRuleFile.m
+                                  tsgWriteMatrix.m)
+
+foreach(Tasmanian_matlab_source_file ${Tasmanian_matlab_source_files})
+    configure_file(${Tasmanian_matlab_source_file} "matlab/${Tasmanian_matlab_source_file}" COPYONLY)
```
dalg24: Changes in the Matlab source files will trigger a re-configure (i.e. re-run CMake and generate the build system again).  Is that what you actually want?

mkstoyanov: Yes, that is the intended behaviour. The use case is:
```
cmake ...
make
octave --eval "addpath('<build dir>/MATLAB/matlab'); tsgCoreTests();'
ERROR ....
```
Check the error; find the buggy script in `<source dir>/InterfaceMATLAB/<script>.m`; edit the script to fix the bug; then I have to retest ...
```
make
octave --eval "addpath('<build dir>/MATLAB/matlab'); tsgCoreTests();'
```

If I use just the `file()` cmake command, the second `make` would not modify the build folder and I would have to rerun cmake manually to update the script.

dalg24: My point was that this can be achieved without re-configuring.  See [this thread](https://stackoverflow.com/questions/8434055/cmake-copy-if-original-file-changed) on stackoverflow.

mkstoyanov: The cmake system is fast, but your point is valid. Unfortunately, the advice on stackoverflow doesn't seem to work right. You can take a look if the problem is obvious:
https://github.com/mkstoyanov/TASMANIAN/tree/cmake_matlab_scripts_update
inside InterfaceMATLAB check CMakeLists.txt and tsgCoreTests.m for the `disp("changed")`

I will try to fix it myself, but this is low priority at the moment.

PR #106, opened 2018-08-07
OpenMP does not work with templates

mkstoyanov:
* OpenMP directives rely on the pre-processor, templates are instantiated after the pre-processor
* will think about this in the future, but OpenMP will make little difference here

### DIFF
--- a/SparseGrids/tsgGridFourier.hpp
+++ b/SparseGrids/tsgGridFourier.hpp
@@ -143,7 +143,6 @@ protected:
         IndexSet *work = (points == 0 ? needed : points);
         int num_points = work->getNumIndexes();
 
-        #pragma omp parallel for
         for (int i=0; i<num_points; i++){
             std::complex<double> basis_entry(1.0,0.0);
             for (int j=0; j<num_dimensions; j++) basis_entry *= cache[j][exponents->getIndex(i)[j] + middles[j]];


dalg24:
What do you mean it "does not work with templates"?

mkstoyanov:
@dalg24 you cannot put OpenMP directives inside of a template, or you need to do some special magic, I am still investigating a bit.
OpenMP `#pragma omp` are handled by the pre-processor, but the templates are instantiated later in the compile process. Thus, I am getting `unknown pragma` warning. I will investigate further, but there is a good argument that OpenMP doesn't belong in this specific function anyway.

dalg24:
> you cannot put OpenMP directives inside of a template

Yes you can.

```C++
#include <vector>
template <typename T>
void foo(std::vector<T> &v)
{
int const n = v.size();
#pragma omp for
for (int i = 0; i < n; ++i)
  v[i] += static_cast<T>(i);
}
int main()
{
std::vector<double> v(100, 3.14);
foo(v);
}
```
This compiled without warning and ran fine for me.

mkstoyanov:
Ouch, I was totally off on that one! :<

The problem is not in OpenMP and templates, the problem is with template and OpenMP directive in the header file which gets included by targets that do not use OpenMP ...

PR #110, opened 2018-08-09
Reformat Fourier coefficients

zbmorrow:
This branch reformats the internal storage of `fourier_coefs`. Internally, the coefficients are arranged so that all real parts are in the first block, and all imaginary parts are in the second block. In CLI-write, `fourier_coefs` is written to a file in the usual interwoven format. MATLAB, Python, and CLI-print will report the coefficients to the user as complex numbers.

(There were no edits to the MATLAB files because they already assumed that CLI-write uses the interwoven format.)


zbmorrow:
I did not have Fortran enabled on my machine when I built the code. Working to fix the issues...

PR #164, opened 2018-09-21
Small cmake updates

mkstoyanov:
* cmake now checks if python ctypes is present (before the check was only for numpy)
* removed unnecessary `import sys`, not needed when using `ENVsetup.sh`
* the package-config now defines targets `Tasmanian_libsparsegrid` and `Tasmanian_libdream` without the `_shared` or `_static` suffixes, thus the user no longer has to be explicit in the type of library


mkstoyanov:
* cmake now checks if python ctypes is present (before the check was only for numpy)
* removed unnecessary `import sys`, not needed when using `ENVsetup.sh`
* the package-config now defines targets `Tasmanian_libsparsegrid` and `Tasmanian_libdream` without the `_shared` or `_static` suffixes, thus the user no longer has to be explicit in the type of library

PR #176, opened 2018-09-27
Fixes for Intel compiler

dalg24:


InterfaceFortran/TasmanianSG.f90
```diff
@@ -685,7 +685,7 @@ function tsgGetComplexHierarchicalCoefficients(gridID) result(c)
     allocate(c_real(2*tsgGetNumOutputs(gridID)*tsgGetNumPoints(gridID)))
     call tsgghc(gridID, c_real)
     do i = 1,size(c)
-      c(i) = complex( c_real(i), c_real(i + size(c)) )
+      c(i) = cmplx( c_real(i), c_real(i + size(c)) )
```
mkstoyanov: https://software.intel.com/en-us/node/679123
`If kind is present, the kind parameter is that specified by kind; otherwise, the kind parameter is that of default real type.`
You need to have:
`c(i) = cmplx( c_real(i), c_real(i + size(c)), 8 )`

InterfaceFortran/TasmanianSG.f90
```diff
@@ -702,7 +702,7 @@ subroutine tsgGetComplexHierarchicalCoefficientsStatic(gridID, c)
   if ( tsgIsFourier(gridID) ) then
     call tsgghc(gridID, c_real)
     do i = 1,size(c)
-      c(i) = complex( c_real(i), c_real(i + size(c)) )
+      c(i) = cmplx( c_real(i), c_real(i + size(c)) )
```
mkstoyanov: `c(i) = cmplx( c_real(i), c_real(i + size(c)), 8 )`

see comment before, this defaults to single-precision

InterfaceFortran/CMakeLists.txt
```diff
@@ -44,8 +44,8 @@ endmacro()
 add_executable(Tasmanian_fortester                fortester.f90)
 add_executable(Tasmanian_example_sparse_grids_f90 Examples/example_sparse_grids.f90)
 
-set_target_properties(Tasmanian_fortester                PROPERTIES OUTPUT_NAME "fortester")
-set_target_properties(Tasmanian_example_sparse_grids_f90 PROPERTIES OUTPUT_NAME "example_sparse_grids_f90")
+set_target_properties(Tasmanian_fortester                PROPERTIES OUTPUT_NAME "fortester" LINKER_LANGUAGE Fortran)
```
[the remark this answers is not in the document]
mkstoyanov: * There are two resolutions:
* If Intel + newer cmake work without explicit `LINKER_LANGUAGE`, than just put in the manual that `ifort` requires newer cmake
* Second options:
```
set_target_properties(Tasmanian_fortester                PROPERTIES OUTPUT_NAME "fortester")
set_target_properties(Tasmanian_example_sparse_grids_f90 PROPERTIES OUTPUT_NAME "example_sparse_grids_f90")
if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Intel")
    set_target_properties(Tasmanian_fortester                PROPERTIES LINKER_LANGUAGE Fortran)
    set_target_properties(Tasmanian_example_sparse_grids_f90 PROPERTIES LINKER_LANGUAGE Fortran)
endif()
```
* The second option is acceptable only if it will not interfere with Intel + CUDA (at the moment CUDA is higher priority than Fortran90/95)

* I can make the changes if you don't want to deal with it, just let me know if newer cmake work with Intel.
* Did the earlier C++ bug get resolved with my latest updates to ORNL master?

PR #185, opened 2018-10-01
Final small fixes

mkstoyanov:
* fixed some last minute bugs
* silenced development debug I/O in some tests
* updated some documentation

### DIFF
--- a/Config/AltBuildSystems/testTSG.py
+++ b/Config/AltBuildSystems/testTSG.py
@@ -146,13 +146,17 @@ class TestTasmanian(unittest.TestCase):
         print("                Available GPUs:")
         if (grid.getNumGPUs() > 0):
             for iGPU in range(grid.getNumGPUs()):
-                sName = grid.getGPUName(iGPU)
+                sName = ""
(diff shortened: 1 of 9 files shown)


Config/AltBuildSystems/testTSG.py
```diff
@@ -146,13 +146,17 @@ def testBasicIO(self):
         print("                Available GPUs:")
         if (grid.getNumGPUs() > 0):
             for iGPU in range(grid.getNumGPUs()):
-                sName = grid.getGPUName(iGPU)
+                sName = ""
+                if ("GNU" == "MSVC"):
```
dalg24: ????

mkstoyanov: The file is automatically generated from `if (@CMAKE_CXX_COMPILER_ID@ == "MSVC")`

Had to disable one of the tests in Visual C++, passing null terminated strings from C++ to Python was crashing, but it could have been something to do with CUDA. This is only the test script, not the actual Python module. I wouldn't put that in the module.

PR #189, opened 2018-10-18
Version bump short

mkstoyanov:
* added support for CUDA 10
* removed the deprecated MS Windows build system with batch scripts
* require cmake 3.10 or newer
    * removed the clumsy work-around for OpenMP on old cmake systems
    * removed other small legacy cmake fixes


mkstoyanov:
@dalg24: I get "404" from Jenkins, when I click on "Details" to see why this is failing. Logging directly into Jenkins doesn't help, I cannot see the details or console output from the test.

CMakeLists.txt
```diff
@@ -1,8 +1,7 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.10.2)
```
dalg24: Why not just `3.10`?

mkstoyanov: I have not tested `3.10.1`, it is possible that there is something wrong. Also, I read somewhere, that cmake adds features to the patches, not just bug-fixes.

PR #191, opened 2018-10-25
Lambda index manipulation

mkstoyanov:
* converted the Sequence grid multi-index sets to RAII memory managements
* created a new `MultiIndexSet` class to replace the old Index, Granulated, and Unsorted sets
* updated algorithms for index selection based on templates and lambdas
* updated the tests that combine level limits and minimum points growth
* *note:* this is a necessary step to update the dynamic refinement algorithms


mkstoyanov:
@dalg24 I removed the redundant calls to `copy()` and `move()`, which improved readability. I also fixed the pointer issue, and added some comments. I call it "ready to merge", unless you have a suggestion.

PR #225, opened 2019-02-14
updating cxx standard to the new cmake way

mkstoyanov:
* C++11 standard is set in a new way in CMake post 3.8 (old way doesn't propagate transitively)


mkstoyanov:
@dalg24 The C++ standard used to be set through `CXX_STANDARD`, but it did not propagate transitively. The CMake post 3.8 method uses `target_compile_features()` which does propagate. Tasmanian 6.0 was using CMake 3.5, I could not have done this before. Got it in develop.

PR #245, opened 2019-03-25
cuda enables as a cmake language

mkstoyanov:
* cuda is enabled as a cmake language
* added all the required hacks to avoid extraneous dependencies and feature regressions

### DIFF
--- a/Config/CMakeIncludes/FindTasmanianCudaMathLibs.cmake
+++ b/Config/CMakeIncludes/FindTasmanianCudaMathLibs.cmake
@@ -1,0 +1,55 @@
+########################################################################
+# FindTasmanianCudaMathLibs.cmake module
+########################################################################
+#
+# requires: must be called AFTER enable_language(CUDA)
+#
+# Uses the CUDA compiler location to deduce the location of
+# the cuBlas and cuSparse Math libraries.
+# The CUDA Run-time libraries are also located for fallback purposes.
+
+# the macro is called as Tasmanian_find_cuda_libraries(NAMES foo1 foo2 PREFIX bar LIST saloon)
+# which will search for foo1 and foo2 using bar as hints, then put the search result in
+# variables called Tasmanian_foo1 and Tasmanian_foo2
+# finally, all results will be appended into a list called Tasmanian_saloon
+macro(Tasmanian_find_cuda_libraries)
+# it is tempting to say NO_DEFAULT_PATH here, but newer versions of CUDA
+# install cuBlas in /usr/lib while nvcc and the other libs sit in /usr/local/cuda
+    cmake_parse_arguments(Tasmanian_cudalibs "" "PREFIX;LIST" "NAMES" ${ARGN} )
+    foreach(_lib ${Tasmanian_cudalibs_NAMES})
+        find_library(Tasmanian_${_lib} ${_lib}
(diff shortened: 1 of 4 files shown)


mkstoyanov:
@dalg24 I finally figured out how to hack CUDA as a language, but I am not sure if it is even worth the trouble.

Config/CMakeIncludes/FindTasmanianCudaMathLibs.cmake
```diff
@@ -0,0 +1,65 @@
...
+Tasmanian_cuda_search_lib(Tasmanian_cuda   cuda)
+Tasmanian_cuda_search_lib(Tasmanian_cudart cudart)
+
+list(APPEND Tasmanian_cudaruntime   ${Tasmanian_cuda}
+                                    ${Tasmanian_cudart})
+
+list(FILTER Tasmanian_cudaruntime EXCLUDE REGEX "-NOTFOUND$")
+
+message(STATUS "Tasmanian found CUDA math libraries: ${Tasmanian_cudamathlibs}")
+message(STATUS "Tasmanian found CUDA runtime libraries: ${Tasmanian_cudaruntime}")
```
dalg24: So we just gather all the ones that are available we don't do something like
```CMake
find_package_handle_standard_args(TasmanianCudaMathLibs DEFAULT_MSG
  Tasmanian_cusparse_static Tasmanian_cublas_static ...)
```

Config/CMakeIncludes/FindTasmanianCudaMathLibs.cmake
```diff
@@ -0,0 +1,65 @@
...
+                 HINTS "${Tasmanian_nvccroot}/lib/"
+                 HINTS "${Tasmanian_nvccroot}/${CMAKE_LIBRARY_ARCHITECTURE}/lib/"
+                 HINTS "${Tasmanian_nvccroot}/lib/${CMAKE_LIBRARY_ARCHITECTURE}"
+                 HINTS "${Tasmanian_nvccroot}/lib64/"
+                 HINTS "${Tasmanian_nvccroot}/${CMAKE_LIBRARY_ARCHITECTURE}/lib64/"
+                 HINTS "${Tasmanian_nvccroot}/lib64/${CMAKE_LIBRARY_ARCHITECTURE}")
+endmacro(Tasmanian_cuda_search_lib)
+
+get_filename_component(Tasmanian_nvccroot ${CMAKE_CUDA_COMPILER} DIRECTORY)
+get_filename_component(Tasmanian_nvccroot ${Tasmanian_nvccroot} DIRECTORY)
```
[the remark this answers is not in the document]
mkstoyanov: That's so much better. I had not dabbled in macro so much before. Thanks!

Config/CMakeIncludes/FindTasmanianCudaMathLibs.cmake
mkstoyanov: Not all libraries are required, different versions of cuda use different combination of auxiliary static/Lt/device libraries. The main cublas and cusparse are required, the rest ... I assume what's needed is present, otherwise I have to build a database for each cuda version.

I used the `find_package_handle_standard_args` for cublas and cusparse, to handle the `Required` libs. But the macro only prints the first library, so I will keep the verbose message for now. May suppress it for the release, but it is good for debugging for now.

PR #254, opened 2019-03-30
Movable main class

mkstoyanov:
* added move constructor to TasmanianSparseGrid class
* add move and assignment operator=


mkstoyanov:
@dalg24 There was a copy constructor a while back, had to add a move constructor; could not figure out how to reuse the constructor for the operator=.

mkstoyanov:
@dalg24 Never mind, I should have simply used `= default;`

PR #258, opened 2019-04-03
Update random engine

mkstoyanov:
* switched sparse grid testing to implementation independent random number generator
* updated the dream random seed selection

### DIFF
--- a/DREAM/tasdreamExternalTests.cpp
+++ b/DREAM/tasdreamExternalTests.cpp
@@ -112,13 +112,12 @@ bool DreamExternalTester::compareSamples(const std::vector<double> &lower, const
 bool DreamExternalTester::testGaussian3D(){
     bool passAll = true;
     int num_dimensions = 3;
-    int rseed = 42, num_samples = 1000, num_chains = 20;
+    int num_samples = 1000, num_chains = 20;
     int num_iterations = num_samples / num_chains + 2;
     int num_burnup = 20 * num_iterations;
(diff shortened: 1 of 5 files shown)


mkstoyanov:
@dalg24 I am getting this failed build, but only gibberish when I try to get to the ctest console output. All tests pass on my machine with and without docker.

dalg24:
> @dalg24 I am getting this failed build, but only gibberish when I try to get to the ctest console output. All tests pass on my machine with and without docker.

Just saw what got printed on the [standard output](https://cloud.cees.ornl.gov/jenkins-ci/job/Tasmanian-GCC54-CUDA90/343/testReport/projectroot/SparseGrids/SparseGridsAcceleration/)  and yes that's not very helpful.  I am not sure what the issue is.  How urgent is this?  Do you have access to a machine that has nvidia-docker?

mkstoyanov:
It is not too urgent.

I have nvidia-docker installed on my workstation, but the problem doesn't appear here. In fact, the PR passes all nightly tests. 

Could this be the case of someone else using the GPUs and running out of RAM? The test that is failing is the C++/CUDA test with larger problems, the Python/CUDA test passed and those run smaller problems.

mkstoyanov:
I'll merge the PR, it will go through the nightly tests, I can revert it tomorrow if I get a report for an error.

`ERROR: cuda failed at CudaVector::resize(), call to cudaMalloc() with error: out of memory`

PR #292, opened 2019-06-06
getPointsIndexes() for Fourier grids

zbmorrow:
small patch: add case for Fourier grids

### DIFF
--- a/SparseGrids/TasmanianSparseGrid.cpp
+++ b/SparseGrids/TasmanianSparseGrid.cpp
@@ -1219,18 +1219,10 @@ const double* TasmanianSparseGrid::getHierarchicalCoefficients() const{
     }
 }
 const int* TasmanianSparseGrid::getPointsIndexes() const{
-    if (isLocalPolynomial()){
-        return getGridLocalPolynomial()->getPointIndexes();
-    }else if (isWavelet()){
-        return getGridWavelet()->getPointIndexes();
-    }else if (isGlobal()){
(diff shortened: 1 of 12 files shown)


SparseGrids/TasmanianSparseGrid.cpp
```diff
@@ -1227,6 +1227,8 @@ const int* TasmanianSparseGrid::getPointsIndexes() const{
         return getGridGlobal()->getPointIndexes();
     }else if (isSequence()){
         return getGridSequence()->getPointIndexes();
+    }else if (isFourier()){
```
mkstoyanov: This works, but we have to make it right.
* Go to `tsgGridCore.hpp` and add the method:
```
const int* GridLocalPolynomial::getPointIndexes() const{
    return ((points.empty()) ? needed.getIndex(0) : points.getIndex(0));
}
```
* delete the corresponding method from `tsgGrid<type>` for all types (Global, Sequence, Localp, Wavelet)
* from `TasmanianSparseGrid.hpp` call only
```
if (empty())
 throw error ...
else
 base->getPointsIndexes();
```

mkstoyanov: The reason is that there are separate methods in separate classes doing essentially the same, we can unify those into a single common place and common call.

mkstoyanov: Good!

PR #296, opened 2019-06-13
begin anisotropic refinement for Fourier grids

zbmorrow:
- start base functionality for `estimateAnisotropicCoefficients()` for Fourier grids
- highly similar to `tsgGridSequence.cpp` and `tsgGridGlobal.cpp`
- add tests for `estimateAnisotropicCoefficients` in Fourier case

### DIFF
--- a/SparseGrids/tasgridExternalTests.cpp
+++ b/SparseGrids/tasgridExternalTests.cpp
@@ -36,6 +36,20 @@
 
 std::minstd_rand park_miller(10);
 
+void loadValues(const BaseFunction *f, TasmanianSparseGrid &grid){
+    int num_dimensions = grid.getNumDimensions();
+    int num_outputs    = grid.getNumOutputs();
+    int num_needed     = grid.getNumNeeded();
+    if (num_needed > 0){
+        std::vector<double> points;
+        grid.getNeededPoints(points);
+        std::vector<double> values(num_outputs * num_needed);
+        for(int i=0; i<num_needed; i++)
+            f->eval(&points[i*num_dimensions], &values[i*num_outputs]);
+        grid.loadNeededPoints(values);
+    }
+}
+
 ExternalTester::ExternalTester(int in_num_mc) : num_mc(in_num_mc), verbose(false), gpuid(-1) {}
 ExternalTester::~ExternalTester(){}
 void ExternalTester::resetRandomSeed(){ park_miller.seed(static_cast<long unsigned>(std::time(nullptr))); }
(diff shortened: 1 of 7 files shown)


SparseGrids/tsgGridFourier.cpp
```diff
@@ -518,7 +523,94 @@ void GridFourier::clearAccelerationData(){
...
+    for(int i=0; i<num_points; i++) n += (max_fcoef[i] > tol) ? 1 : 0;
+
+    Data2D<double> A(n,m);
+    std::vector<double> b(n);
+
+    int count = 0;
+    for(int c=0; c<num_points; c++){
+        const int *indx = points.getIndex(c);
+        if (max_fcoef[c] > tol){
+            for(int j=0; j<num_dimensions; j++) A.getStrip(j)[count] = ((type == type_hyperbolic) || (type == type_iphyperbolic) || (type == type_qphyperbolic)) ? log((double) (indx[j]+1)) : ((double) indx[j]);
```
mkstoyanov: * Did you look at the condition number and rank of the matrix here? The log term is quite unstable when I use it in the Global Grid setting, but it could be simply because of the mix of linear and log terms.
* Don't use the tripple-or, use
```
bool ishyperbolic = (OneDimensionalMeta::getControurType(type) == type_hyperbolic);
for(int j=0; j<num_dimensions; j++) A.getStrip(j)[count] = (ishyperbolic) ? log((double) (indx[j]+1)) : ((double) indx[j]);
```

zbmorrow: The condition number of A (with the log-term) for a 2D grid with L=5 is about 9.

mkstoyanov: That's good. Hopefully it will work right.

SparseGrids/tsgGridFourier.cpp
```diff
@@ -518,7 +523,94 @@ void GridFourier::clearAccelerationData(){
     cuda_cache.reset();
     #endif
 }
-void GridFourier::clearRefinement(){ return; }     // to be expanded later
+
+void GridFourier::estimateAnisotropicCoefficients(TypeDepth type, int output, std::vector<int> &weights) const{
```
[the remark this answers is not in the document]
zbmorrow: Tests added in latest commit.

SparseGrids/tasgridExternalTests.cpp
```diff
@@ -1357,7 +1357,39 @@ bool ExternalTester::testAllRefinement() const{
     }
     cout << "      Construction              dynamic/local" << setw(15) << ((pass4) ? "Pass" : "FAIL") << endl;
 
-    return (pass && pass2 && pass3 && pass4);
+    bool pass5 = true;
+    {
+        const BaseFunction *f = &f21c1c2periodic;
```
mkstoyanov: If the function is C^1 - C^2 then how come the anisotropy is (3, 4) as opposed to (2, 3).

zbmorrow: Functions in C^k (i.e., continuous and periodic derivatives up to order k) have Fourier coefficients c_n that decay like O(n^-(k+2)). So we should have alpha=(1+2, 2+2)=(3, 4).

zbmorrow:
I've moved around the Fourier anisotropy test and used `loadValues()`. Are further changes needed?

PR #297, opened 2019-06-17
Refinement strategy

zbmorrow:
This pull request adds the functions and data structures that will be used in the refinement strategy. All functions are currently empty.


SparseGrids/tsgGridFourier.hpp
```diff
@@ -183,6 +189,11 @@ class GridFourier : public BaseCanonicalGrid {
     MultiIndexSet tensors;
     MultiIndexSet active_tensors;
     std::vector<int> active_w;
+
+    MultiIndexSet updated_tensors;
```
mkstoyanov: Leave the variables out of this
* having unused variables will generate an unnecessary warning
* the variables should be added to the file I/O, (make another PR with just variables and file I/O)

zbmorrow: Done

PR #313, opened 2019-07-03
add mpi to the docker image

mkstoyanov:
* Makes sure that we add MPI to the docker image the next time we update the image. 
* MPI testing is enabled in the nightly builds, so actual updating can wait.


mkstoyanov:
* I build a very similar image for nightly testing, I will let you know when I make a docker hub account
* I will not need MPI + CUDA for now, I will set it for the nightly build before I dabble with something so arcane in the CI
* I hope this is not interfering with your vacation

PR #364, opened 2019-08-14
Python addons

mkstoyanov:
* here is a working example of the python callback using `numpy.ndarrays` and multi-threading
* the most important files are `loadNeededPoints.cpp`, `TasmanianAddons.in.py` and the example 5

### DIFF
--- a/Addons/tsgCLoadNeededValues.cpp
+++ b/Addons/tsgCLoadNeededValues.cpp
@@ -35,6 +35,7 @@
 extern "C"{
 
 void tsgLoadNeededPoints(){
+    std::cout << "Hello" << std::endl;
 }
 
 } // extern "C"
--- a/Addons/tsgCLoadNeededValues.cpp
+++ b/Addons/tsgCLoadNeededValues.cpp
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2017, Miroslav Stoyanov
+ *
+ * This file is part of
+ * Toolkit for Adaptive Stochastic Modeling And Non-Intrusive ApproximatioN: TASMANIAN
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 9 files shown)


InterfacePython/example_sparse_grids_05.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+        sInfo += "{0:>14s}".format("")
+        print(sInfo)
+        bBelowBudget = (grid_isotropic.getNumLoaded() < iBudget
+                       or grid_icurved.getNumLoaded() < iBudget
+                       or grid_icurved.getNumLoaded() < iBudget
+                       or grid_surplus.getNumLoaded() < iBudget)
+
+        bBelowBudget = (grid_isotropic.getNumLoaded() < iBudget)
```
dalg24: Did you mean this 2nd assignment?

mkstoyanov: I am still debugging this, without the second assignment it will cause an infinite loop. It won't  when I add all other grids.

Addons/tsgCLoadNeededValues.cpp
```diff
@@ -0,0 +1,70 @@
...
+// x.size(), x.data(), y.size(), y.data(), thread_id
+using tsg_lnp_model = void (*)(int, double const*, int, double*, int);
+
+void tsgLoadNeededPoints(int overwrite, tsg_lnp_model pymodel, void *grid_pntr, int num_threads){
+    TasGrid::TasmanianSparseGrid &grid = *reinterpret_cast<TasGrid::TasmanianSparseGrid*>(grid_pntr);
+
+    int num_dimensions = grid.getNumDimensions();
+    int num_outputs    = grid.getNumOutputs();
+
+    auto cpp_model = [&](double const x[], double y[], size_t thread_id)->
```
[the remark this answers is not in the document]
mkstoyanov: It captures the two ints and the function pointer, neither changes so I don't see a point to assign.

I can change the into to `const` to make it more clear.

PR #425, opened 2019-10-07
Automatic build of documentation on merge

dalg24:


.jenkins
```diff
@@ -0,0 +1,28 @@
...
+                }
+            }
+            environment {
+                GIT_CREDS = credentials('0f170146-bd5e-44c0-a6e0-4f6fac3230fa')
+            }
+            steps {
+                sh 'apt update && apt install --no-install-recommends -y build-essential cmake doxygen git'
+                sh 'git config http.sslVerify false'
+                sh '''rm -rf build_doc && mkdir -p build_doc && cd build_doc &&
+                      cmake -D Tasmanian_ENABLE_DOXYGEN=ON .. && make Tasmanian_doxygen &&
```
mkstoyanov: Minor point, here we should add `-D DOXYGEN_INTERNAL_DOCS=ON`
```
cmake -D DOXYGEN_INTERNAL_DOCS=ON -D Tasmanian_ENABLE_DOXYGEN=ON; ...
```

dalg24: Done

.jenkins
```diff
@@ -0,0 +1,35 @@
...
+            }
+            steps {
+                sh 'apt update && apt install --no-install-recommends -y build-essential cmake doxygen git'
+                sh '''rm -rf build_doc && mkdir -p build_doc && cd build_doc &&
+                      cmake -DTasmanian_ENABLE_DOXYGEN=ON -DDOXYGEN_INTERNAL_DOCS=ON .. &&
+                      make Tasmanian_doxygen &&
+                      git fetch origin gh-pages && git branch -D gh-pages &&
+                      git checkout -b gh-pages FETCH_HEAD &&
+                      rm -rf ../rtd && mv Doxygen/html ../rtd && git add --all ../rtd &&
+                      git config http.sslVerify false
```
dalg24: I didn't manage to get it to work without that...

.jenkins
```diff
@@ -0,0 +1,35 @@
...
+                GIT_CREDS = credentials('0f170146-bd5e-44c0-a6e0-4f6fac3230fa')
+            }
+            steps {
+                sh 'apt update && apt install --no-install-recommends -y build-essential cmake doxygen git'
+                sh '''rm -rf build_doc && mkdir -p build_doc && cd build_doc &&
+                      cmake -DTasmanian_ENABLE_DOXYGEN=ON -DDOXYGEN_INTERNAL_DOCS=ON .. &&
+                      make Tasmanian_doxygen &&
+                      git fetch origin gh-pages && git branch -D gh-pages &&
+                      git checkout -b gh-pages FETCH_HEAD &&
+                      rm -rf ../rtd && mv Doxygen/html ../rtd && git add --all ../rtd &&
```
dalg24: Note that this will publish under https://ornl.github.io/TASMANIAN/rtd/
We can symlink rtfd to rtd :)

.jenkins
```diff
@@ -0,0 +1,35 @@
+pipeline {
+    agent {
+        label 'DoNotRunAnythingOnMaster'
+    }
+    stages {
+        stage("Deploy documentation") {
```
dalg24: I would like to migrate the other builds to this jenkinsfile.  I plan on doing that later if you agree that it is better this way.

.jenkins
mkstoyanov: Why delete the branch and why the subfolder? I guess I'm missing something, why not
```
# assuming we start at master and we have build added to .gitignore
mkdir -p build; cd build
cmake -D DOXYGEN_INTERNAL_DOCS=ON -D Tasmanian_ENABLE_DOXYGEN=ON ..
make Tasmanian_doxygen
cd ..; git checkout gh-pages; # checkout gh-pages then delete the current page
git rm -r *.map *.md5 *.png *.html search/ *.js *.css formula.repository
cp -r build/Doxygen/html/* .
git add *
git commit -m "new page" # added the new page
git reset --soft HEAD^^; git commit -m "updated page" # squash the last two commits
git push -f                                                                       # force push
```
That way `gh-pages` has no source, only the web-page, and the `index.html` is at the root folder so the address is:  (https://ornl.github.io/)[https://ornl.github.io/]

We can add the `RTFD` on the title on the first page :)

.jenkins
```diff
@@ -0,0 +1,35 @@
...
+                docker {
+                    image "ubuntu:18.04"
+                    label 'docker'
+                }
+            }
+            environment {
+                GIT_CREDS = credentials('0f170146-bd5e-44c0-a6e0-4f6fac3230fa')
+            }
+            steps {
+                sh 'apt update && apt install --no-install-recommends -y build-essential cmake doxygen git'
```
mkstoyanov: We also need `graphviz` to get the Doxygen dot module for the formulas.

Also, are we rebuilding the docker image on every pull-request? Seems slow and wasteful.

.jenkins
mkstoyanov: Are we rebuilding the images on every pull request? Otherwise I'm OK with the move.

mkstoyanov: Are we rebuilding the images on every pull request? Otherwise I'm OK with the move.

.jenkins
dalg24: I would like to have the option to use the root page (https://ornl.github.io/TASMANIAN/) for something else, possibly a home page.

.jenkins
dalg24: Images will be cached and rebuilt if something changed.  That's how we do it on ArborX and Kokkos now.

.jenkins
dalg24: I was too lazy to define a Dockerfile but I will add it

dalg24: Done.

.jenkins
mkstoyanov: OK. I would like to use `/rolling/` for this page, in the future we can have `/stable/` for the latest stable release (or even `/stableX.Y/` although we wouldn't have more than one).

Right now both stable and the home-page are under `tasmanian.ornl.gov` which is good for as long as we can keep it.

PR #437, opened 2019-10-14
Include git commit hash in documentation header

dalg24:
Tentative fix for https://github.com/ORNL/TASMANIAN/issues/417#issuecomment-541971636


Doxygen/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ set(DOXYGEN_HTML_COLORSTYLE_GAMMA  "90")
 set(DOXYGEN_FORMULA_FONTSIZE "14")
 
 set(DOXYGEN_PROJECT_NAME   "Toolkit for Adaptive Stochastic Modeling and Non-Intrusive ApproximatioN: ${PROJECT_NAME} v${Tasmanian_VERSION_MAJOR}.${Tasmanian_VERSION_MINOR}${Tasmanian_version_comment}")
-set(DOXYGEN_PROJECT_NUMBER "") # if project number is given, the rendered font is too small
+set(DOXYGEN_PROJECT_NUMBER "${Tasmanian_git_hash}") # if project number is given, the rendered font is too small
```
mkstoyanov: FYI: I don't like this, doesn't look good at all. If you want the hash on-line, I would put it somewhere else, e.g., after the page link:
```
Documentation: development (rolling, currently at hash 98ae1315be4d64925284e5b28075e769f5db46fb)
```

dalg24: I am fine with that.  Also I think `Tasmanian_git_hash` was not defined because I can't see it on the doc.

mkstoyanov: See #438

mkstoyanov:
Addressed by #438

PR #439, opened 2019-10-15
added texlive and ghostscript to doxygen

mkstoyanov:
* the formulas also require `latex` and `ghostscript`, added to the docker


mkstoyanov:
Yes, I build the container locally and the formulas showed up.

Issue #417, opened 2019-09-30
Add autodeployment for the Doxygen web-page


mkstoyanov:
* currently the web-page in `gh-pages` is updated manually
* add an automated procedure to rebuild the pages upon merging a pull request

dalg24:
Partially addressed by #425 and #432

dalg24:
Remaining issue if the documentation generated is the same as the one that is already online:
```
On branch gh-pages
Untracked files:
	./

nothing added to commit but untracked files present
```
Full log [here](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/TASMANIAN/detail/master/8/pipeline/)

mkstoyanov:
We should also squash the commits, one the process is polished, no need to keep 2 - 3MB per PR.

mkstoyanov:
With #438 and #439 the process is officially working. Thanks @dalg24

mkstoyanov:
Minor adjustments left:
* the web-page build on merge, but the regular tests run again
    * do not repeat the CI when the PR is merged
* the gh-pages branch will keep growing, especially due to binary files such as formulas and charts
    * squash the last two commits and force-push to the gh-pages

mkstoyanov:
squashing commits is addressed in #441

mkstoyanov:
It works and it's alive! https://ornl.github.io/TASMANIAN/

Thanks, @dalg24 !

PR #441, opened 2019-10-16
Squash doxygen commits

mkstoyanov:
* using squash of commits, only keep the last doxygen page

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -134,12 +134,12 @@ cmake \
                       make Tasmanian_doxygen &&
                       git config http.sslVerify false
                       git fetch origin gh-pages &&
-                      git checkout -B gh-pages FETCH_HEAD &&
+                      git checkout -B gh-pages FETCH_HEAD^ &&
                       rm -rf ../rolling && mv Doxygen/html ../rolling && git add --all ../rolling &&
                       git config user.email "noreply@ornl.gov" &&
                       git config user.name "Jenkins CI" &&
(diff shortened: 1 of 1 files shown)


.jenkins
```diff
@@ -134,7 +134,9 @@ cmake \
                       git config user.email "noreply@ornl.gov" &&
                       git config user.name "Jenkins CI" &&
                       git commit -m "Automated documentation update" &&
-                      git push https://$GIT_CREDS@github.com/ORNL/TASMANIAN.git gh-pages'''
+                      git reset --soft HEAD^^ &&
```
dalg24: Why 2x ^?

mkstoyanov: If I understand the logic:
* move head to the last commit that should not be changed
* take everything after that commit and merge it into a single commit
Using only one ^ will redo the last commit, i.e., no change at all.

mkstoyanov: I've tested it.

mkstoyanov:
Addresses a point in #417

dalg24:
I am still confused with the two `git commit`

mkstoyanov:
The first is a commit, the second (when combined with the soft-reset) is a `rebase -i`

I'll see if I can do that with only one commit.

mkstoyanov:
I got it much simpler and I don't think it gets any simpler.
* when checking out, checkout all but the last commit
* commit the new page
* force-push which will overwrite the previous last commit

PR #485, opened 2020-02-20
Add Fortran 2003 wrappers

sethrj:
@mkstoyanov Rather than investing in the infrastructure up-front (it shouldn't be very difficult, in truth) I thought I'd demonstrate the automated wrapping input and generated code.


mkstoyanov:
This answers a lot of questions, but brings up more.

1. I'm having hard time to see the user-code, what is the Fortran equivalent of the following 3 lines:
```
grid.makeGlobalGrid(2, 1, 3, TasGrid::type_level, TasGrid::rule_clenshawcurtis);
std::vector<double> grid.getPoints();
```
2. Many calls to Tasmanian use both default variables and vector-array overloads.
    * we want Fortran to support the default variables
    * we don't need both arrays (int *) and vectors (std::vector<int>) as those do the same job
3. Another note on vectors, I see `_wrap_VecInt_front_ref` which has nothing to do with Tasmanian but std::vector. Will there be an issue if two projects both use vectors and have swig interfaces, wouldn't there be a conflict between the vector methods.

sethrj:
@mkstoyanov In this example I instructed SWIG to include wrappers for `std::vector` since a lot of the sparse grid interface used it. My last push deletes the vector wrappers and ignores all the methods that use `std::vector`.

The following code (or something very similar) should be how to create a grid and load points into a 2D fortran array:
```fortran
use, intrinsic :: ISO_C_BINDING
use tasmanian
type(TasmanianSparseGrid) :: grid
real(C_DOUBLE), dimension(:,:), allocatable :: points

grid = TasmanianSparseGrid()
call grid%makeGlobalGrid(2, 1, 3, type_level, rule_clenshawcurtis)
allocate(points(grid%getNumDimensions(), grid%getNumPoints()))
call grid%getPoints(reshape(points, [size(points)]))
```

mkstoyanov:
@sethrj I like this! We should do this, it can even sit along the current Fortran 90 API (although that API should go away eventually, users permitting).

There are minor naming things like prefixing the enums with `tsg_` and giving the module a proper name, do you want to continue the discussion here or meet face-to-face?

sethrj:
Excellent! The enum and module naming are easy fixes. Let's meet in person, maybe tomorrow? With probably an hour's work I think we can get this integrated into your build system and get the high-level behavior tuned to your project's needs.

mkstoyanov:
@mtlaiu Do you think we can use this within Tornado? Might be better than the current setup, will be easier to maintain for sure.

mtlaiu:
I don't see why we can't. I can test both approaches with Thornado and let you know if any differences observed.

mkstoyanov:
@mtlaiu You shouldn't see any difference in performance, the question is whether they will be OK with the Fortran 2003 and object oriented syntax.

mtlaiu:
I think it shouldn't be a problem but I'll need to doublecheck with them on this.

InterfaceSwig/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+  endif()
+endfunction()
+
+########################################################################
+# Fortran librareis and command line tools
+########################################################################
+
+set(Tasmanian_libfor_target_name tasmanian)
+tasmanian_add_swig_module(${Tasmanian_libfor_target_name})
+target_link_libraries(${Tasmanian_libfor_target_name} Tasmanian_addons)
```
mkstoyanov: This line needs to change to:
```
target_link_libraries(${Tasmanian_libfor_target_name} Tasmanian_addons Tasmanian_libdream_${Tasmanian_lib_default})
```
Should fix the current issue with the CI, and I would do it myself but I have no permission to push to the branch.

sethrj: @mkstoyanov I granted you access to my fork. Alternatively you can create your own branch using
```sh
git fetch upstream pull/485/head:swig
git checkout swig
```

InterfaceSwig/generated/tasmanian.f03
```diff
@@ -0,0 +1,3848 @@
...
+import :: swigclasswrapper
+integer(C_INT), intent(in) :: farg1
+integer(C_INT), intent(in) :: farg2
+integer(C_INT), intent(in) :: farg3
+integer(C_INT), intent(in) :: farg4
+type(SwigClassWrapper) :: fresult
+end function
+
+function swigc_makeWaveletGrid__SWIG_2(farg1, farg2, farg3) &
+bind(C, name="_wrap_ma
```
mkstoyanov: I got to the point where I can generate the files locally, but I am getting errors when I try to compile.

Gfortran and I don't think this is a valid Fortran declaration.

InterfaceSwig/generated/tasmanian.f03
```diff
@@ -0,0 +1,3848 @@
...
+  procedure :: evaluateSparseHierarchicalFunctionsGetNZ => swigf_TasmanianSparseGrid_evaluateSparseHierarchicalFunct1CQM1I
+  procedure :: evaluateSparseHierarchicalFunctionsStatic => swigf_TasmanianSparseGrid_evaluateSparseHierarchicalFunct1HRJEE
+  procedure :: getPointsIndexes => swigf_TasmanianSparseGrid_getPointsIndexes
+  procedure :: getNeededIndexes => swigf_TasmanianSparseGrid_getNeededIndexes
+  procedure, private :: swigf_TasmanianSparseGrid_op_assign__
+  generic :: setDomainTransform => swigf_TasmanianSparseGrid_setDomainTransform__SWIG_1
+  generic :: getDomainTransform => swigf_TasmanianSparseGrid_getDomainTransform__SWIG_1
+  generic :: setAnisotropicRefinement => swigf_TasmanianSparseGrid_setAnisotropicRefinement__SWIG_1, &
+    swigf_TasmanianSparseGrid_setAnisotropicRefinement__SWIG_2
+  generic :: setSurplusRefinement => swigf_TasmanianSparseGrid_setSurplusRefinement__SWIG_1, &
```
mkstoyanov: I got to the point where I can generate the files locally, but I am getting errors when I try to compile.

Gfortran claims this is ambiguous, probably because _SWIG_1 and _SWIG_7 have the same signature. On the C++ side, the difference is between an int and an enum.

InterfaceSwig/generated/tasmanian.f03
mkstoyanov: Github seems to be confusing the lines in a large file, the line with bad code is:
```
integer(C_INT), dimension(getPointsIndexes*) :: swig_result
```

sethrj: Aha, I know what's causing that: my unrestricted use of the `%apply` typemap to treat pointers as arrays isn't working correctly for the %typemap(out). I'll need to create a minimal test case and add to the `swig-fortran` issues page to track it.

InterfaceSwig/generated/tasmanian.f03
sethrj: Aha, that makes sense. Is the C++ class overloaded between an int and enum? We'll have to instruct SWIG to `%ignore` one of the methods (passing in the full signature).

mkstoyanov: We cannot `%ignore` since the methods do different things, the int and the enum are not equivalent. In other interfaces, where overloads are not well supported, I split the overload family into two sets.

sethrj: Well, we can `%rename` one set of methods. So if I'm unfolding the overloads and default arguments correctly (and ignoring the additional overloads with `std::vector`, we have:
```
void setSurplusRefinement(double, int);
void setSurplusRefinement(double, int, const int *);
void setSurplusRefinement(double, TypeRefinement);
void setSurplusRefinement(double, TypeRefinement, int);
void setSurplusRefinement(double, TypeRefinement, int, const int *);
void setSurplusRefinement(double, TypeRefinement, int, const int *, const double *);
```

I think you can resolve this with:
```swig
%rename(setSurplusRefinementOutput) void TasGrid::TasmanianSparseGrid::setSurplusRefinement(double, int);
%rename(setSurplusRefinementOutput) void TasGrid::TasmanianSparseGrid::setSurplusRefinement(double, int, const int *);
```

mkstoyanov: To avoid confusion, lets do the following:

Without an enum, the methods refer to `Global` version of the surplus refinement:
```
%rename(setGlobalSurplusRefinement) void TasGrid::TasmanianSparseGrid::setSurplusRefinement(double, int);
%rename(setGlobalSurplusRefinement) void TasGrid::TasmanianSparseGrid::setSurplusRefinement(double, int, const int *);
```

And for the enum refers specifically to a Local sub-strategy for surplus refinement:
```
%rename(setLocalSurplusRefinement) void TasGrid::TasmanianSparseGrid::setSurplusRefinement(double, TypeRefinement);
%rename(setLocalSurplusRefinement) void TasGrid::TasmanianSparseGrid::setSurplusRefinement(double, TypeRefinement, int);
%rename(setLocalSurplusRefinement) void TasGrid::TasmanianSparseGrid::setSurplusRefinement(double, TypeRefinement, int, const int *);
%rename(setLocalSurplusRefinement) void TasGrid::TasmanianSparseGrid::setSurplusRefinement(double, TypeRefinement, int, const int *, const double *);
```

sethrj: Might be worth changing the C++ interface itself in that case, actually... I did not immediately see that set of functions was global and the other local.

mkstoyanov: You might be right, I have to think some more.

The enum criteria specifies one of the refinement variations, but global grids (defined by previous use of makeGlobalGrid()) are very rigid and do not allow for variations. 

The rest of the inputs are the same and this is one of the few cases where the underlying grid type matters for interface purposes.

mkstoyanov: @sethrj You had a point on the C++ interface, I will make a change.

For now, we can just ignore the overloads without enums. I will do that, I need to get my feet wet a bit with the swift code anyway. I'm also in the middle of making other changes, so don't change anything on your side.

sethrj: Great! Let me know if you need me to point you in the right direction on anything SWIG-related.

mtlaiu:
> @mtlaiu You shouldn't see any difference in performance, the question is whether they will be OK with the Fortran 2003 and object oriented syntax.
@mkstoyanov  I checked with them and they confirmed that the Fortran compilers on the machine should be able to handle most of the object oriented syntax in Fortran 2003. Performance is a prtential issue but, as you mentioned, it shouldn't matter to us as we are using it as a wrapper.

mkstoyanov:
@mtlaiu this is good news since the new interface will be much easier to maintain

We can keep the existing one (since it is already there) and the new one (since we also need to do some tests) and we can make a full switch only when we are comfortable with the results. The savings in maintenance will come in the long run.

As for performance, the old and new wrappers work more or less the same, but the new interface may have a bit of an advantage since the bind method added in 2003 translates the Fortran method call to a C method call at compile time, while the old interface made a Fortran and a C call.

mkstoyanov:
@sethrj I got things to compile with swig and without swig (so long as the module is pre-generated). I figured out how to ignore the problematic functions, I we don't need those anyway.

The problem is that I cannot seem to write the basic test. See the fortranswigtester.f90 file. The grid is generated and the sizes are correct, but the getPoints method is not working. Maybe I am calling it wrong with the reshape.

sethrj:
@mkstoyanov Excellent work! I was able to get the example to work by directly passing a 1D array to SWIG and then reshaping the result. I'm not sure what is wrong with the `reshape`: I'm either misunderstanding the reference/value semantics of `reshape`, or the mechanism of assuming the reshaped data is `stride-1` is failing. (This is actually the only real standards non-compliant part of the SWIG interface layer: assuming an array of length N can be represented by the pointer to the first element, and the size.)

sethrj:
Yep it's my misunderstanding of `reshape` based on my use of Numpy. In numpy, `reshape` returns a view to the original array. In Fortran, it apparently returns a *copy*. You can check that the `reshape` I originally suggested (which caused your test to fail) is wrong by trying to pass the `reshape`d array to a procedure with a corresponding `intent(out)` argument. It'll say:
```fortran
   56 |     call set_my_points(2, 5, reshape(temp_points, shape(points)))
      |                             1
Error: Non-variable expression in variable definition context (actual argument to INTENT = OUT/INOUT) at (1)
```

mkstoyanov:
I don't like the reshape-copy syntax, I found out that this works:
```
real(C_DOUBLE), dimension(:,:), allocatable :: points

allocate(points(grid%getNumDimensions(), grid%getNumPoints()))
call grid%getNeededPoints(points(:, 1))
```
I wrote a few tests and it seems OK.

Overall, I like that the interface mimics the C++ so closely and it can be used by just following the C++ reference.

mkstoyanov:
I would like to fix the Swig warnings about the TasmanianSparseGrid move constructor and the reference and pointer overloads of copyGrid(), but I somehow failed to add the right syntax.

Even without fixing the warnings, I am happy to merge.

sethrj:
@mkstoyanov I fixed the warnings. Glad it's worked out so well!

mkstoyanov:
@sethrj Beautiful!

There are some enhancements that I would like to make later on, but I will bug you about it in the future. This stage is successfully complete.

sethrj:
@mkstoyanov BTW, regarding the `.f90` vs `.f03` name extensions: https://software.intel.com/en-us/blogs/2013/01/11/doctor-fortran-in-source-form-just-wants-to-be-free
Apparently the "best practice" is to use `.f90` to denote the "free-form", and not the fortran standard compatibility.

mkstoyanov:
Yeah, I had to figure that out the hard way. The current version is using .f90 for all Fortran files, just different folders and the final compiled library has different name.

Thanks for the note, and let's move further Fortran discussions to issue #492 as this PR is now merged.

PR #562, opened 2021-01-21
for parallel evaluations in matlab

zbmorrow:
Add process ID for evaluating the interpolant with Matlab parallel toolbox

### DIFF
--- a/InterfaceMATLAB/tsgMakeFilenames.m
+++ b/InterfaceMATLAB/tsgMakeFilenames.m
@@ -7,13 +7,19 @@ function [sFileG, sFileX, sFileV, sFileO, sFileW, sFileC, sFileL] = tsgMakeFilen
 
 [sFiles, sTasGrid] = tsgGetPaths();
 
+if exist('getpid','builtin')
+    id = getpid();
+else
+    id = feature('getpid');
+end
+
 sFileG = [sFiles, sGridName,'_FileG']; % the filename to store the grid
 
-sFileX = [sFiles, sGridName,'_FileX']; % file with points to evaluate the surrogate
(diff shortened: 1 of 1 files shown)


InterfaceMATLAB/tsgMakeFilenames.m
```diff
@@ -7,13 +7,15 @@
 
 [sFiles, sTasGrid] = tsgGetPaths();
 
+id = feature('getpid');
```
mkstoyanov, line 10: Looks good and I don't mind the idea, but there are several notes:
* the CI doesn't test if all the files have been deleted after use, double-check to make sure the temporary files are not left in the work folder
* does this work under Octave and MATLAB both?

Note that batching the points into a single call to Tasmanian and using parallelism on the C++ side will always be faster, unless you are doing different things in the MATLAB threads and synchronization is too expensive. You don't want to use parallel MATLAB just for the purpose of calling tagEvaluate().

Comment back, when you can double-checked the points above.

mkstoyanov (changes requested):

InterfaceMATLAB/tsgMakeFilenames.m
zbmorrow: * Turns out MATLAB and Octave have different commands for this. I'll update the pull request shortly.
* Verified that the temp files are deleted post-eval (on my Mac at least) for both MATLAB and Octave

Yes, ordinarily I'd use the C++ parallelism, but our MD code sets a parallel flag within an `fsolve` command that calls the interpolant.

mkstoyanov (approved):

InterfaceMATLAB/tsgMakeFilenames.m
mkstoyanov: Looks good!

PR #563, opened 2021-01-28
GitHub Actions CI for Tasmanian

vreshniak:
This PR adds CI testing for the ORNL/TASMANIAN repo.

One file is added:
1. ```.github/workflows/build-ubuntu.yaml``` performs tests on GitHub hosted Linux runner. It is set to be triggered automatically on push/pull and tests several versions of ubuntu, compiler, and python

One file is modified:
1. ```README.md``` adds badge for the build status


.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  build:
+    runs-on: ${{ matrix.runner }}
+    continue-on-error: false
+    strategy:
+      matrix:
+        runner: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
```
mkstoyanov, line 23: Are we really running 12 tests here, seems excessive :)

Skip Ubuntu 16.04, that one is effectively dead and it would require different cmake since it comes with 3.5 and we require 3.10.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,59 @@
...
+        compiler: [clang++, g++]
+        python: [python2, python3]
```
mkstoyanov, line 25: Skip Python 2 too, effectively it is dead (or should be dead) so there's no point to use it here.

This CI will complement the other one, there is no need to overlap testing too much. And we need to keep the other CI for the GPU capabilities.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,59 @@
...
+          sudo apt install --no-install-recommends -y
+          build-essential
+          clang
+          python
+          python-numpy
+          python3
+          python3-numpy
+          cmake
+          gfortran
+          git
```
mkstoyanov, line 40: We don't need git. Git is only needed to update the web-page.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,59 @@
...
+          libopenmpi-dev
```
mkstoyanov, line 41: We don't need MPI either, we don't test it below and it can be skipped for this CI. The `openmpi-bin` and `openssh-client` are also part of MPI.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,59 @@
...
+          -D CMAKE_INSTALL_PREFIX=./TasmanianInstall
+          -D CMAKE_CXX_FLAGS="-O3 -Wall -Wextra -Wshadow -pedantic"
+          -D CMAKE_CXX_COMPILER=${{ matrix.compiler }}
+          -D Tasmanian_ENABLE_PYTHON=ON
+          -D PYTHON_EXECUTABLE=/usr/bin/${{ matrix.python }}
+          -D Tasmanian_ENABLE_FORTRAN=ON
+          -D Tasmanian_TESTS_OMP_NUM_THREADS=4 .. &&
+          make -j4
+      - name: Test
+        run: cd build && ctest -j4 -V --no-compress-output -T Test
```
mkstoyanov, line 57: `ctest -j4` will launch 4 tests simultaneously while `Tasmanian_TESTS_OMP_NUM_THREADS=4` will launch 4 threads per test. That's total of 16 threads.

Just making sure this is what you intended.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Install
+        run: cd build && make install && make test_install
```
mkstoyanov, line 59: Add a new line here to remove the red mark.

README.md
```diff
@@ -1,3 +1,5 @@
+![build](https://github.com/ORNL/TASMANIAN/workflows/build-ubuntu/badge.svg?branch=master)
```
mkstoyanov, line 1: Do we need this one here?

The documentation and python-pip scripts read from the README.md and make temporary changes, I would have to double-check that we are not interfering with those.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,59 @@
...
+        python: [python2, python3]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install requirements
+        run: >
+          sudo apt install --no-install-recommends -y
+          build-essential
+          clang
+          python
```
mkstoyanov, line 34: Skip Python 2, see the comment above.

.github/workflows/build-ubuntu.yaml
```diff
@@ -0,0 +1,59 @@
...
+        runner: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
+        compiler: [clang++, g++]
+        python: [python2, python3]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install requirements
+        run: >
+          sudo apt install --no-install-recommends -y
+          build-essential
```
mkstoyanov, line 32: Interestingly `build-essentials` doesn't bring in `g++`, just `gcc`. I think `clang` brings in `g++` as a dependence, but don't be surprised if it doesn't and you see errors.

mkstoyanov (changes requested):
Thanks for doing this!

Squash your commits, there's no point to add 30 commits 60 lines of code. Also, feel free to Teams me.

README.md
mkstoyanov: I'll need to make couple of changes after we merge this, but it will be OK.

mkstoyanov (approved):

PR #601, opened 2021-08-10
Added wrappers for LAPACK tridiagonal eigenvalue finders.

wwkong:


InterfaceTPL/tplUnitTests.cpp
```diff
@@ -0,0 +1,139 @@
...
+    } else if (test == unit_none) {
+        cout << "No tests to run.\n";
+        return true;
+    } else {
+        cout << "Unknown test type!\n";
+        return false;
+    }
+}
+
+bool testTriToeplitzEigen() {
```
mkstoyanov, line 75: Move this to `tasgridUnitTests.cpp` and add a new tests. You can use the existing `doesMatch()` methods.

InterfaceTPL/tplUnitTests.cpp
```diff
@@ -0,0 +1,139 @@
...
+    // Initialize.
+    const int N = 100;
+    bool is_matched;
+    double exact_eigs[N], D[N], E[N-1], W[N], Z[N*N], WORK1[2 * N - 2], WORK2[4 * N];
```
mkstoyanov, line 79: Do not allocate large arrays on the stack, use the heap with `std::vector`.

InterfaceTPL/tplUnitTests.cpp
```diff
@@ -0,0 +1,139 @@
...
+    int IBLOCK1[N], ISPLIT1[N], IWORK1[3 * N];
+    double a = M_PI_2;
+    double b = M_PI_4;
+    ISPLIT1[0] = N;
+    IBLOCK1[0] = N;
+    for (int i=0; i<N; i++) {
+        exact_eigs[i] = a + 2.0 * b * cos((i+1) * M_PI / (N+1));
```
mkstoyanov, line 86: The MS compiler doesn't like `M_PI`, use `TasGrid::Maths::pi` defined in `tsgMathUtils.hpp`.

Also, use `std::cos()`

InterfaceTPL/tplUnitTests.cpp
```diff
@@ -0,0 +1,139 @@
...
+    std::fill_n(E, N-1, b);
+    TasBLAS::steqr('N', N, D, E, Z, 1, WORK1);
+    std::sort(D, D + N);
+    is_matched = doesMatch(N, D, exact_eigs);
+
+    // Test LAPACK's dstebz function.
+    std::fill_n(D, N, a);
+    std::fill_n(E, N-1, b);
+    TasBLAS::stebz('A', 'E', N, 0.0, 0.0, 1, N, 1e-13, D, E, N, 1, W, IBLOCK1, ISPLIT1, WORK2, IWORK1);
+    is_matched = doesMatch(N, W, exact_eigs);
```
mkstoyanov, line 108: What would happen if the call to `TasBLAS::steqr()` on line 100 fails, but the call to `TasBLAS::stebz()` on line 107 succeeds? 

If the QR method fails, then the unit test should fail, right?

InterfaceTPL/tplUnitTests.cpp
wwkong: Done.

InterfaceTPL/tplUnitTests.cpp
wwkong: Done, except I used arrays instead of std::vectors.

InterfaceTPL/tplUnitTests.cpp
wwkong: Yes. This logic has been fixed.

SparseGrids/CMakeLists.txt
```diff
@@ -56,7 +56,6 @@ set(Tasmanian_source_libsparsegrid_cuda
        ${CMAKE_CURRENT_SOURCE_DIR}/../InterfaceTPL/tsgCudaWrappers.cpp
        tsgCudaKernels.cu)
 
-
```
mkstoyanov, line 59: Why remove the line?

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -34,6 +34,8 @@
 #include "tasgridUnitTests.hpp"
 #include "tasgridExternalTests.hpp"
 #include "tasgridTestHelpers.hpp"
+#include "tsgBlasWrappers.hpp"
+#include "tsgMathUtils.hpp"
```
mkstoyanov, line 38: It is transitively included through TasmanianSparseGrid.hpp, no need to include it here.

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -34,6 +34,8 @@
 #include "tasgridUnitTests.hpp"
 #include "tasgridExternalTests.hpp"
 #include "tasgridTestHelpers.hpp"
+#include "tsgBlasWrappers.hpp"
```
mkstoyanov, line 37: Somewhere in the BLAS wrappers, there should be pre-processor clause that guards in the case when BLAS is not enabled. Make sure it works with and without BLAS.

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -45,6 +47,15 @@
 
 #ifdef Tasmanian_ENABLE_DPCPP
 #include <CL/sycl.hpp>
+#include <string>
```
mkstoyanov, line 50: Lots of includes that are included transitively from `tsgEnumerates.hpp` and visible pretty much all throughout the code.

Also, be careful when adding code. This block sits inside `#ifdef Tasmanian_ENABLE_DPCPP` when it has nothing to do with DPC++.

SparseGrids/CMakeLists.txt
wwkong: I added some modifications there and deleted them, but forgot to add back the newline. It should be fixed now.

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -786,4 +800,69 @@ std::vector<std::function<void(void)>> GridUnitTester::getRuntimeErrorCalls() co
     };
 }
 
+bool GridUnitTester::testLAPACKInterface() {
+
+    bool all_matched = true;
+    #ifdef Tasmanian_ENABLE_BLAS
+    // Initialize.
+    const int N = 100;
+    double* exact_eigs = new double[N];
```
mkstoyanov, line 809: NO!

Use RAII and `std::vector`. The clauses for `new` and `delete` should only ever be used in very VERY special circumstances. If they can be avoided, then they should always be avoided.

Watch this:
https://www.youtube.com/watch?v=7Qgd9B1KuMQ

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -786,4 +800,69 @@ std::vector<std::function<void(void)>> GridUnitTester::getRuntimeErrorCalls() co
...
+
+    // Test LAPACK's dstebz function.
+    std::fill_n(D, N, a);
+    std::fill_n(E, N-1, b);
+    TasBLAS::stebz('A', 'E', N, 0.0, 0.0, 1, N, 1e-13, D, E, N, 1, W, IBLOCK1, ISPLIT1, WORK2, IWORK1);
+    all_matched = all_matched && doesMatch(N, W, exact_eigs);
+
+    // For debugging only.
+    // for (int i=0; i<N; i++) {
+    //     cout << std::to_string(D[i]) << " " <<  std::to_string(W[i]) << " " << std::to_string(exact_eigs[i])<< endl;
```
mkstoyanov, line 850: No need to convert to string when passing into a stream, the `<<` operator will take care of this.

SparseGrids/tasgridUnitTests.hpp
```diff
@@ -32,9 +32,11 @@
 #define __TASGRID_UNIT_TESTS_HPP
 
 #include "tasgridCLICommon.hpp"
+#include <string>
+#include <vector>
 
-enum UnitTests{
-    unit_none, unit_all, unit_cover, unit_except, unit_api, unit_c
+enum UnitTests {
```
mkstoyanov, line 38: Why add the space, there is no need to change this line.

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -786,4 +800,69 @@ std::vector<std::function<void(void)>> GridUnitTester::getRuntimeErrorCalls() co
...
+    // Test LAPACK's dsterf function.
+    std::fill_n(D, N, a);
+    std::fill_n(E, N-1, b);
+    TasBLAS::sterf(N, D, E);
+    std::sort(D, D + N);
+    all_matched = all_matched && doesMatch(N, D, exact_eigs);
+
+    // Test LAPACK's dsteqr function.
+    std::fill_n(D, N, a);
+    std::fill_n(E, N-1, b);
```
mkstoyanov, line 837: Are `D` and `E` destroyed in the previous call. Why do you need to initialize again?

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -786,4 +800,69 @@ std::vector<std::function<void(void)>> GridUnitTester::getRuntimeErrorCalls() co
...
+    double b = TasGrid::Maths::pi / 4;
+    ISPLIT1[0] = N;
+    IBLOCK1[0] = N;
+    for (int i=0; i<N; i++) {
+        exact_eigs[i] = a + 2.0 * b * std::cos((i+1) * TasGrid::Maths::pi / (N+1));
+    }
+    std::sort(exact_eigs, exact_eigs + N);
+
+    // Test LAPACK's dsterf function.
+    std::fill_n(D, N, a);
```
mkstoyanov, line 829: Switching the above to `std::vector` will allow you to initialize the vector when you create it, then there is no need to call the `fill()` method.

InterfaceTPL/tsgBlasWrappers.hpp
```diff
@@ -196,7 +201,75 @@ namespace TasBLAS{
                 throw std::runtime_error(std::string("Lapack zgels_ infer-worksize-stage exited with code: ") + std::to_string(info));
         }
     }
-    #ifdef Tasmanian_BLAS_HAS_ZGELQ
+    /*!
+     * \ingroup TasmanianTPLWrappers
+     * \brief LAPACK dstebz
+     */
+    inline void stebz(char range, char order, int N, double vl, double vu, int il, int iu, double abstol, double D[], double E[],
+                      int M, int nsplit, double W[], int iblock[], int isplit[], double work[], int iwork[]) {
```
mkstoyanov, line 209: There is an inconsistency here. `M` is an output of the eigensolver, but you are passing in by value so the C++ wrapper will not return the value.

mkstoyanov (changes requested):

SparseGrids/tasgridUnitTests.cpp
wwkong: According to the LAPACK documentation (http://www.netlib.org/lapack/explore-html/d2/d24/group__aux_o_t_h_e_rcomputational_gaf0616552c11358ae8298d0ac18ac023c.html), D is both an input (diagonal entries) and output (computed eigenvalues) while the contents of E are destroyed in the previous call.

SparseGrids/tasgridUnitTests.hpp
wwkong: Fixed.

SparseGrids/tasgridUnitTests.cpp
mkstoyanov: Note that the person who starts the conversation should be the one to hit the "resolve" button. Once the conversation is resolved, github hides it and I cannot see your answer unless I find it and click to unhide.

Otherwise, you're right here and you have to refill the vector, but the first fill can still be done in the initialization stage. Also, make sure that the C++ wrappers properly reflect the const or non-const aspect of a variable.

wwkong: Gotcha. I'll keep all future conversations open.

SparseGrids/tasgridUnitTests.cpp
wwkong: Removed in the latest commit.

SparseGrids/tasgridUnitTests.cpp
wwkong: Wouldn't it be better to use pre-processor directives in tasgridUnitTest.cpp (see the latest commit) itself?

SparseGrids/tasgridUnitTests.cpp
wwkong: Removed all redundant includes in the latest commit.

SparseGrids/tasgridUnitTests.cpp
wwkong: I also tested the latest commit with Tasmanian_ENABLE_BLAS=OFF and it seems to be working fine.

SparseGrids/tasgridUnitTests.cpp
wwkong: Done.

SparseGrids/tasgridUnitTests.cpp
wwkong: Changed.

SparseGrids/tasgridUnitTests.cpp
wwkong: Done. I initialized D and E appropriately and removed the ``std::fill_n()``.

InterfaceTPL/tsgBlasWrappers.hpp
wwkong: Changed. I also made the same correction for ``nsplit``.

SparseGrids/tasgridUnitTests.cpp
mkstoyanov: The main point is that this should work with `Tasmanian_ENABLE_BLAS` and don't want the C++ wrappers to be visible when they refer to LAPACK functions that cannot be linked. Look at `tsgTPLWrappers.hpp` and perhaps you should include that file.

InterfaceTPL/tsgBlasWrappers.hpp
```diff
@@ -73,6 +73,11 @@ void dgels_(const char *trans, const int *M, const int *N, const int *nrhs, doub
             double *B, const int *ldb, double *work, int *lwork, int *info);
 void zgels_(const char *trans, const int *M, const int *N, const int *nrhs, std::complex<double> *A, const int *lda,
             std::complex<double> *B, const int *ldb, std::complex<double> *work, int *lwork, int *info);
+// Symmetric tridiagonal eigenvalue compute
+void dstebz_(const char *range, const char *order, const int *N, const double *vl, const double *vu, const int *il, const int *iu, const double *abstol,
+             const double D[], const double E[], int *M, int *nsplit, double W[], int iblock[], int isplit[], double work[], int iwork[], int *info);
```
mkstoyanov, line 78: In an earlier comment you mentioned that `E` is destroyed during the call. Is that so? If so, then this should not be `const double E[]`. Double check the other arrays.

SparseGrids/CMakeLists.txt
```diff
@@ -180,6 +180,10 @@ add_test(SparseGridsAPI          gridtest api)
 add_test(SparseGridsC            gridtest c)
 Tasmanian_set_test_properties(TESTS SparseGridsAcceleration SparseGridsDomain SparseGridsRefinement SparseGridsGlobal SparseGridsLocal SparseGridsWavelet SparseGridsFourier SparseGridsExceptions SparseGridsAPI SparseGridsC)
 
+if (Tasmanian_ENABLE_BLAS)
+	add_test(SparseGridsLAPACK gridtest lapack)
+    Tasmanian_set_test_properties(TESTS SparseGridsLAPACK)
```
mkstoyanov, line 185: Is this aligned to 4 spaces like the rest of the code? We use 4 spaces to align in Tasmanian.

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -786,4 +793,57 @@ std::vector<std::function<void(void)>> GridUnitTester::getRuntimeErrorCalls() co
     };
 }
 
+bool GridUnitTester::testLAPACKInterface() {
+    bool all_matched = true;
+
+    #ifdef Tasmanian_ENABLE_BLAS
```
mkstoyanov, line 799: Adding the pre-processor here is good. Also make sure that the BLAS header is not included unless BLAS is declared.

SparseGrids/tasgridUnitTests.cpp
mkstoyanov: See the comment below about `#ifdef Tasmanian_ENABLE_BLAS`.

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -786,4 +793,57 @@ std::vector<std::function<void(void)>> GridUnitTester::getRuntimeErrorCalls() co
...
+
+    // Test LAPACK's dstebz function.
+    std::fill(D.begin(), D.end(), a);
+    std::fill(E.begin(), E.end(), b);
+    TasBLAS::stebz('A', 'E', N, 0.0, 0.0, 1, N, 1e-13, D.data(), E.data(), M,
+                   nsplit, W.data(), IBLOCK1.data(), ISPLIT1.data(),
+                   WORK2.data(), IWORK1.data());
+    all_matched = all_matched && doesMatch(W, exact_eigs);
+
+    // For debugging only.
```
mkstoyanov, line 840: Get rid of this in the final version.

SparseGrids/tasgridUnitTests.cpp
```diff
@@ -786,4 +793,57 @@ std::vector<std::function<void(void)>> GridUnitTester::getRuntimeErrorCalls() co
...
+    std::vector<double> D(N, a), E(N-1, b);
+    std::vector<double> W(N), Z(N*N), WORK1(2*N-2), WORK2(4*N);
+    std::vector<int> IBLOCK1(N), ISPLIT1(N), IWORK1(3*N);
+    ISPLIT1[0] = N;
+    IBLOCK1[0] = N;
+
+    // Test LAPACK's dsterf function.
+    TasBLAS::sterf(N, D.data(), E.data());
+    std::sort(D.begin(), D.end());
+    all_matched = all_matched && doesMatch(D, exact_eigs);
```
mkstoyanov, line 823: This is correct, but if the test fails you wouldn't know which of the three methods fail.

Add an if-statement:
```
if (not doesMatch(...)){
    std::cout << "ERROR: failed LAPACK test at sterf()\n";
   all_match = false;
}
```

mkstoyanov (changes requested):

SparseGrids/tasgridUnitTests.cpp
wwkong: Done.

SparseGrids/CMakeLists.txt
wwkong: Changed.

InterfaceTPL/tsgBlasWrappers.hpp
wwkong: Only ``dbsterf`` and ``dsteqr`` destroy ``E``. The function ``dstebz``, on the other hand, outputs the eigenvalues to another array ``W``.

For reference, see:
http://www.netlib.org/lapack/explore-html/d2/d24/group__aux_o_t_h_e_rcomputational_gaf0616552c11358ae8298d0ac18ac023c.html
http://www.netlib.org/lapack/explore-html/d2/d24/group__aux_o_t_h_e_rcomputational_ga47fe470e7a882c58d4cc49e2c6cb7c70.html
https://www.netlib.org/lapack/explore-html/d2/d24/group__aux_o_t_h_e_rcomputational_ga28f88843da09a0ee400daf46caaabec6.html

SparseGrids/tasgridUnitTests.cpp
wwkong: Done. I replaced the conditional import of ``tsgBlasWrappers.hpp`` with an import of ``tsgTPLWrapper.hpp``.

InterfaceTPL/tsgBlasWrappers.hpp
mkstoyanov: OK, so long as everything is consistent.

SparseGrids/tasgridUnitTests.cpp
wwkong: Gotcha. This is handled with ``import tsgTPLWrappers.hpp`` now.

SparseGrids/tasgridUnitTests.cpp
wwkong: Done.

mkstoyanov (approved):

PR #602, opened 2021-08-18
Exotic Quadrature

wwkong:
Development of the one-dimensional exotic quadrature rules.


SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -33,6 +33,7 @@
 
 #include "tsgCoreOneDimensional.hpp"
 #include "tsgIOHelpers.hpp"
+#include "tsgTPLWrappers.hpp"
```
mkstoyanov, line 36: Do not use TPL Wrapper directly, add an eigenvalue method to the tsgLinearSolvers.hpp and tsgLinearSolvers.cpp files.

SparseGrids/tasgridExternalTests.cpp
```diff
@@ -2234,6 +2235,17 @@ bool ExternalTester::testAllAcceleration() const{
 void ExternalTester::debugTest(){
     cout << "Debug Test (callable from the CMake build folder)" << endl;
     cout << "Put testing code here and call with ./SparseGrids/gridtester debug" << endl;
+    int n = 200;
+    int nref = 501;
+    std::function<double(double)> sinc_plus1 =
```
mkstoyanov, line 2240: `std::function` is compatible with a lambda, but it is not really a lambda.

For these cases, you can just use auto:
```
auto sinc_pluis1 = [](double x)-> ....
```

SparseGrids/tasgridExternalTests.cpp
```diff
@@ -34,6 +34,7 @@
 
 #include "tasgridExternalTests.hpp"
 #include "tasgridTestHelpers.hpp"
+#include "tsgCoreOneDimensional.hpp"
```
mkstoyanov, line 37: This file should already be visible here. Generally, I want to avoid redundant inclusions which helps keep things organized.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,72 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
     }
 }
 
+// Generate the n roots of the n-th degree orthogonal polynomial.
+// ref_points and ref_points are used in the computation of the integrals that form the elements of the Jacobi matrix.
+double poly_eval(int nroots, std::vector<double> roots, double x) {
```
mkstoyanov, line 398: Looking at this, is `nroots` different from `roots.size()`?

The vector keeps it's own size, which makes it easy to pass around.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,72 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+        int n,
+        std::function<double(double)> &weight_fn,
+        const std::vector<double> &ref_weights,
+        const std::vector<double> &ref_points) {
+
+    // Initialize and compute the integral weights corresponding to weight_fn.
+    std::vector<std::vector<double>> root_cache(n);
+    std::vector<double> roots;
+    #ifdef Tasmanian_ENABLE_BLAS
+    if (ref_points.size() != ref_weights.size()) {
```
mkstoyanov, line 415: This is a check for the internals, we can just use `assert()`:

```
assert(ref_points.size() == ref_weights.size());
```
The purpose of the exceptions is that the user receives feedback on what went wrong, we don't need to be this verbose in our own internal error checking.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,72 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+std::vector<std::vector<double>> OneDimensionalOrthPolynomials::getRootCache(
+        int n,
+        std::function<double(double)> &weight_fn,
+        const std::vector<double> &ref_weights,
+        const std::vector<double> &ref_points) {
+
+    // Initialize and compute the integral weights corresponding to weight_fn.
+    std::vector<std::vector<double>> root_cache(n);
+    std::vector<double> roots;
+    #ifdef Tasmanian_ENABLE_BLAS
```
mkstoyanov, line 414: Part of the reason for using `tsgLinearAlgebra.hpp` is that the eigenvalue method will be always declared, but the implementation will throw if BLAS is not enabled.

Then you don't need to have pre-processor directives all over the rest of the code.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,72 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+    if (ref_points.size() != ref_weights.size()) {
+        throw std::invalid_argument("The number of reference points does not match the point of reference weights!\n");
+    }
+    int nref = ref_points.size();
```
mkstoyanov, line 418: As a general rule, try using as few variables as possible. It is better to use `ref_points.size();` multiple times than to have an extra variable.

If you have to have a variable like this one, make sure to declare it as `const`.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,72 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+    std::vector<double> alpha(n), beta(n-1), offdiag(n-1);
+    double alpha_numr, alpha_denm, beta_numr, beta_denm;
+    for (int i=0; i<n; i++) {
+        // Form the tridiagonal vectors of the Jacobi matrix.
+        alpha_numr = 0.0;
+        alpha_denm = 0.0;
+        beta_numr = 0.0;
+        beta_denm = 0.0;
+        for (int j=0; j<nref; j++) {
+            alpha_numr += ref_points[j] * pow(poly_vals[j], 2) * integral_weights[j];
```
mkstoyanov, line 434: * use `std::pow()` as opposed to just `pow` which can be in any namespace
* the integer version of `std::pow()` has been deprecated, use `std::pow(..., 2.0)`
    * the compiler will convert 2 to 2.0, but it's better to be explicit
* don't use `std::pow()` at all here, you can just use `poly_vals[j] * poly_vals[j]`
    * I'm worried that `pow()` will use logarithms for potentially non-integer powers which will waste many flops

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,72 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+    }
+    int nref = ref_points.size();
+    std::vector<double> integral_weights(nref);
+    for (int j=0; j<nref; j++) {
+        integral_weights[j] = weight_fn(ref_points[j]) * ref_weights[j];
+    }
+    // Compute the roots incrementally.
+    std::vector<double> poly_m1_vals(nref, 0.0), poly_vals(nref, 1.0);
+    std::vector<double> alpha(n), beta(n-1), offdiag(n-1);
+    double alpha_numr, alpha_denm, beta_numr, beta_denm;
```
mkstoyanov, line 426: In general, do not leave uninitialized variables.

Move the declaration of these inside the for-loop.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,72 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+        roots.resize(i+1);
+        TasBLAS::sterf(i+1, roots.data(), offdiag.data());
+        root_cache[i] = roots;
+        // Update the values of the polynomials at ref_points.
+        poly_m1_vals = poly_vals;
+        std::transform(ref_points.begin(), ref_points.end(), poly_vals.begin(),
+                       [&roots, i](double x){return poly_eval(i+1, roots, x);});
+    }
+    #endif
+    return(root_cache);
```
mkstoyanov, line 459: `return` is not a function, no need for brackets.

SparseGrids/tsgCoreOneDimensional.hpp
```diff
@@ -184,6 +184,18 @@ namespace OneDimensionalMeta{
...
 }
 
+//! \internal
+//! \brief Contains algorithms for generating orthogonal polynomials with respect to a NONNEGATIVE weight function.
+//! \ingroup TasmanianCoreOneDimensional
+namespace OneDimensionalOrthPolynomials{
+    //! \brief Generate all roots up to the n roots of the n-th degree orthogonal polynomial.
+    std::vector<std::vector<double>> getRootCache(
+        int n,
+        std::function<double(double)> &weight_fn,
```
mkstoyanov, line 194: I would pass in by value here. Otherwise, you may run into issues trying to bind a non-const reference.

mkstoyanov:
I added lots of comments, most of these are probably things that you were going to change anyways, but see if you have any questions.

SparseGrids/tasgridExternalTests.cpp
wwkong: Done.

SparseGrids/tasgridExternalTests.cpp
wwkong: Done. I removed this inclusion.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Then was an older version where ``roots`` was preallocated to have ``n`` elements. The new version has ``roots.size() == i+1`` for each level `i` now, so I have removed the need for ``nroots`` above.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Okay. Changed.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: II removed all references to ``nref``.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Changed. ``pow`` is no longer used.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: I moved all of the doubles and ``offdiag`` inside the for-loop, but ``alpha`` and ``beta`` need to be initialized with zero length outside of the for-loop now.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Changed.

SparseGrids/tsgCoreOneDimensional.hpp
wwkong: Changed.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -395,16 +397,15 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
 
 // Generate the n roots of the n-th degree orthogonal polynomial.
 // ref_points and ref_points are used in the computation of the integrals that form the elements of the Jacobi matrix.
-double poly_eval(int nroots, std::vector<double> roots, double x) {
+double poly_eval(std::vector<double> roots, double x) {
```
mkstoyanov, line 399: * do not pass vectors by value, it will create an unnecessary copy
* always pass vectors by reference, unless you actually want to make a copy
* maintain const-correctness, if you don't intend to change the input, then mark it as `const`

Vectors are almost always accepted as `const std::vector<double> &roots`

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,168 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+        std::vector<std::vector<double>> &points_cache,
+        const int n,
+        const double shift,
+        std::function<double(double)> weight_fn,
+        const int nref) {
+
+    // Create the set of points for the first term.
+    std::vector<double> ref_weights(nref), ref_points(nref);
+    TasGrid::OneDimensionalNodes::getGaussLegendre(nref, ref_weights, ref_points);
+    points_cache.resize(n);
```
mkstoyanov, line 487: No need to resize, it will be overwritten on the next line.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,168 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+    double eval_value = 1.0;
+    for (size_t i=0; i<roots.size(); i++) {
+        if (i != idx) {
+            eval_value *= (x - roots[i]) / (roots[idx] - roots[i]);
+        }
+    }
+    return eval_value;
+}
+void OneDimensionalExoticQuad::getExoticGaussLegendreCache(
+        std::vector<std::vector<double>> &weights_cache,
```
mkstoyanov, line 477: All outputs should follow all inputs, inputs come first.

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,168 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+    return eval_value;
+}
+std::vector<std::vector<double>> OneDimensionalExoticQuad::getRootCache(
+        const int n,
+        std::function<double(double)> weight_fn,
+        const std::vector<double> &ref_weights,
+        const std::vector<double> &ref_points) {
+
+    // Initialize and compute the integral weights corresponding to weight_fn.
+    std::vector<std::vector<double>> root_cache(n);
```
mkstoyanov, line 412: rename `roots_cache` to `roots` and remove `roots`

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,168 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+    // Create the set of points for the first term.
+    std::vector<double> ref_weights(nref), ref_points(nref);
+    TasGrid::OneDimensionalNodes::getGaussLegendre(nref, ref_weights, ref_points);
+    points_cache.resize(n);
+    points_cache = OneDimensionalExoticQuad::getRootCache(
+        n, [shift, weight_fn](double x)->double{return (weight_fn(x) + shift);},
+        ref_weights, ref_points);
+
+    // Create the set of weights for the first term.
+    weights_cache.resize(n);
```
mkstoyanov, line 493:
```
weights_cache = std::vector<std::vector<double>>(n);
for(int i=0; i<n; i++){
    std::vector<double> &quad_weights = wights_cache[i];
   // or just use wights_cache[i]
}
```

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,168 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+    for (int i=0; i<n; i++) {
+        std::vector<double> quad_weights(points_cache[i].size(), 0.0);
+        for (size_t j=0; j<points_cache[i].size(); j++) {
+            // Integrate the function l_j(x) * [weight_fn(x) + shift].
+            for (int k=0; k<nref; k++) {
+                quad_weights[j] +=
+                        lagrange_eval(j, points_cache[i], ref_points[k]) *
+                        (weight_fn(ref_points[k]) + shift) *
```
mkstoyanov, line 501: Reduce the calls to `weight_fn()`

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,168 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+        std::vector<double>
+                points(points_cache[i].size()),
+                gl_weights(points_cache[i].size());
+        TasGrid::OneDimensionalNodes::getGaussLegendre(points_cache[i].size(),
+                                                       gl_weights,
+                                                       points);
+        std::vector<double> quad_weights(points_cache[i].size());
+        for (size_t j=0; j<points_cache[i].size(); j++) {
+            quad_weights[j] = -shift * gl_weights[j];
+            if (points.size() % 2 == 1) {
```
mkstoyanov, line 519: Why inside the body of the for loop?

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,168 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+
+    // Create and append the set of points and weights for the second term.
+    for (int i=0; i<n; i++) {
+        std::vector<double>
+                points(points_cache[i].size()),
+                gl_weights(points_cache[i].size());
+        TasGrid::OneDimensionalNodes::getGaussLegendre(points_cache[i].size(),
+                                                       gl_weights,
+                                                       points);
+        std::vector<double> quad_weights(points_cache[i].size());
```
mkstoyanov, line 516: No need for quad_weights, you can just modify gl_weights

SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -392,6 +393,168 @@ const char* OneDimensionalMeta::getHumanString(TypeOneDRule rule){
...
+        }
+        points_cache[i].reserve(points_cache[i].size() + points.size());
+        points_cache[i].insert(points_cache[i].end(), points.begin(), points.end());
+        weights_cache[i].reserve(weights_cache[i].size() + quad_weights.size());
+        weights_cache[i].insert(weights_cache[i].end(), quad_weights.begin(), quad_weights.end());
+    }
+}
+
+// Write Exotic Gauss-Legendre points and weights to an ASCII formatted file.
+void OneDimensionalExoticQuad::writeExoticGaussLegendreCache(
```
mkstoyanov, line 532: No need to write to an ostream.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Done.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Done.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Done.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Changed.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Changed.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Changed.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Changed to use weights_cache[i] now.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Changed.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Moved outside the for-loop.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Changed.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Removed this function.

Addons/tsgExoticQuadrature.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include "tsgCoreOneDimensional.hpp"
+#include "tsgLinearSolvers.hpp"
+#include <cassert>
+
+namespace TasGrid {
+
+// Evaluates a polynomial with roots given by \b roots at the point \b x.
+double poly_eval(const std::vector<double> &roots, double x) {
+    double eval_value = 1.0;
+    for (size_t i=0; i<roots.size(); i++) {
```
mkstoyanov, line 60:
```
for(auto r : roots) eval_value *= (r - x);
```

Addons/tsgExoticQuadrature.cpp
```diff
@@ -0,0 +1,183 @@
...
+        std::vector<double> offdiag(n-1);
+        for (int k=0; k<i; k++) {
+            offdiag[k] = sqrt(beta[k]);
+        }
+        TasmanianTridiagonalSolver::getSymmetricEigenvalues(i+1, roots[i], offdiag);
+        if (roots[i].size() % 2 == 1) {
+            roots[i][(roots[i].size() - 1) / 2] = 0.0; // Zero out the center for stability.
+        }
+        // Update the values of the polynomials at ref_points.
+        poly_m1_vals = poly_vals;
```
mkstoyanov, line 107: use `std::swap()`

Addons/tsgExoticQuadrature.cpp
```diff
@@ -0,0 +1,183 @@
...
+                       [&roots, i](double x){return poly_eval(roots[i], x);});
+    }
+    return roots;
+}
+
+//! \brief Evaluates a Lagrange basis polynomial at a point \b x.
+double lagrange_eval(int idx, const std::vector<double> roots, double x) {
+    double eval_value = 1.0;
+    for (size_t i=0; i<roots.size(); i++) {
+        if (i != idx) {
```
mkstoyanov, line 118: if -statement in a for-loop can be expensive.

Break into two for-loops going to `0 ... idx-1` and `idx+1 ... roots.size()`

Addons/tsgExoticQuadrature.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    // Create and append the set of points and weights for the second term.
+    for (int i=0; i<n; i++) {
+        std::vector<double>
+                points(points_cache[i].size()),
+                correction_weights(points_cache[i].size());
+        TasGrid::OneDimensionalNodes::getGaussLegendre(points_cache[i].size(),
+                                                       correction_weights,
+                                                       points);
+        for (size_t j=0; j<points_cache[i].size(); j++) {
```
mkstoyanov, line 167: for(auto &w : correction_weights) w *= -shift;

Addons/testAddons.cpp
```diff
@@ -53,29 +58,38 @@ int main(int argc, char const **argv){
...
-    cout << std::setw(40) << "Unstructured construction" << std::setw(10) << ((pass) ? "Pass" : "FAIL") << endl;
-    pass_all = pass_all && pass;
-    #else
-    cout << std::setw(40) << "Unstructured construction" << std::setw(10) << "skipping" << endl;
-    gpuid *= 2; // no op to register the use of gpuid
-    #endif
-
-    cout << "\n";
-    if (pass){
+    if (debug) {
```
mkstoyanov, line 61: No need to shift the alignment here, just put
```
if (debug){
    debugTest();
    return 0;
}
```

Addons/testAddons.cpp
```diff
@@ -30,21 +30,29 @@
...
+    cout << "Debug Test (callable from the CMake build folder)" << endl;
+    cout << "Put testing code here and call with ./Addons/addontester debug" << endl;
+    // debugGetRoots();
+    debugGetExoticQuadrature();
+    // debugSincT(100, 0.031353648322695503);
+    // debugSincT(10, 0.32099682841103033);
+    // debugSincT(1, 1.4321357541271255);
+}
 
-    cout << "\n\n";
```
mkstoyanov, line 36: Make sure you don't delete this one.

Addons/tsgExoticQuadrature.cpp
wwkong: Changed.

Addons/tsgExoticQuadrature.cpp
wwkong: Changed.

Addons/testAddons.cpp
wwkong: Changed.

Addons/testAddons.cpp
wwkong: Noted. This line still remains in the latest commit.

Addons/testExoticQuadrature.hpp
```diff
@@ -0,0 +1,223 @@
...
+}
+inline bool testExpMx2_sinc10_shift1() {
+    auto f = [](double x)->double {return std::exp(-x*x);};
+    bool passed = wrapSincTest1D(f, 20, 10.0, 1.0, 0.32099682841103033);
+    if (not passed) {
+        std::cout << "ERROR: Test failed in testExpMx2_sinc10_shift1()" << std::endl;
+    }
+    return passed;
+}
+inline bool testExpMx2_sinc100_shift1() {
```
mkstoyanov, line 138: This is looking much better now.

Right here, you don't need all of those functions, all that those do is define a test function `f`, call `wrapSincTest1D()` and then print an error message of the rest fails. If you want to see the specific method being called, just add a string parameter to `wrapSincTest1D()` and do the cout statement there. Then call the wrapper directly by inlining the lambda expressions:
```
pass = pass and wrapSincTest1D([](double x)->double { return std::exp(-x*x); }, 20, 100.0, 1.0, 0.031353648322695503);
```

Addons/testExoticQuadrature.hpp
```diff
@@ -0,0 +1,223 @@
...
+    passed = passed && testExpMx2_sinc1_shift1();
+    passed = passed && testExpMx2_sinc10_shift1();
+    passed = passed && testExpMx2_sinc100_shift1();
+    return passed;
+}
+
+// Integrates the function f(x, y) * sinc(freq * x) * sinc(freq * y) over
+// [-1, 1] ^ 2  using Exotic quadrature at a specified level and shift. Then,
+// compares this integral against a given exact integral value.
+inline bool wrapSincTest2D(std::function<double(double, double)> f, int level, double freq, double shift, double exact_intgr) {
```
mkstoyanov, line 158: One key distinction of Tasmanian (and Sparse Grid codes) is the dimension is a flexible parameter. 

Can you think of a way to pass the dimension as either a variable or a template parameter?

Addons/tsgExoticQuadrature.hpp
```diff
@@ -0,0 +1,187 @@
...
+            }
+        }
+    }
+
+    // Create and append the set of correction points and weights for the second term only if the shift is nonzero.
+    if (shift != 0.0) {
+        for (int i=0; i<n; i++) {
+            const int init_size = points_cache[i].size();
+            std::vector<double>
+                    correction_points(init_size),
```
mkstoyanov, line 142: No need to initialize the vectors `getGaussLegendre()` will set the correct size.

Addons/testExoticQuadrature.hpp
```diff
@@ -0,0 +1,162 @@
...
+    return passed;
+}
+
+// Test the attributes of the output of getExoticQuadrature().
+inline bool testBasicAttributes() {
+    bool passed = true;
+    const int n = 3;
+    const int nref = 11;
+    const char *descr = "TEST_DESCRIPTION";
+    auto sinc = [](double x) -> double {return (x == 0.0 ? 1.0 : sin(x) / (x));};
```
mkstoyanov, line 64: Just a general note: This is not the correct way to implement the `sinc()` function.

The problem with the `0.0` in the denominator is not just at `x = 0.0` but also for values close to zero. The correct way is to pick a small range around `0.0` and then use a truncated Maclaurin expansion. There's no need to use that here since the Gauss-Legendre quadrature points will fall away from `0.0`.

Addons/testExoticQuadrature.hpp
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+    return passed;
+}
+
+
+// Given a dimension, integrates the function f(x[1],...,x[dimension]) * sinc(freq * x[1]) * ... * sinc(freq * x[dimension]) over
+// [-1, 1]^dimension  using Exotic quadrature at a given depth and a specified shift. Then, compares this integral against a given exact
+// integral value.
+inline bool wrapSincTest(std::function<double(std::vector<double>)> f, int dimension, int depth, double freq, double shift,
```
mkstoyanov, line 92: This causes two unnecessary copy operations, first when you move the data from the `quad_points` vector to `x` and then when you pass `x` into the function.

The solution in the other Tasmanian tests is to pass a cost pointer `const double*` which can point to the middle of `quad_points` without a copy.

Addons/testExoticQuadrature.hpp
```diff
@@ -0,0 +1,162 @@
...
+    TasGrid::TasmanianSparseGrid sg;
+    sg.makeGlobalGrid(dimension, 1, depth, TasGrid::type_qptotal, std::move(ct));
+
+    // Compute the integral and compare to the reference value.
+    std::vector<double> quad_points = sg.getPoints();
+    std::vector<double> quad_weights = sg.getQuadratureWeights();
+    double approx_integral = 0.0;
+    assert(quad_weights.size() * dimension == quad_points.size());
+    for (size_t i=0; i<quad_weights.size(); i++) {
+        std::vector<double> x(dimension);
```
mkstoyanov, line 109: If you need to copy, you should use the proper constructor:
```
std::vector<double> x(&quad_points[dimension * i], &quad_points[dimension * i] + dimension);
```

Addons/testExoticQuadrature.hpp
```diff
@@ -0,0 +1,162 @@
...
+        for (int d=0; d<dimension; d++) {
+            x[d] = quad_points[dimension * i + d];
+        }
+        approx_integral += f(x) * quad_weights[i];
+    }
+    if (std::abs(approx_integral - exact_integral) > precision) {
+        std::cout << "ERROR: " << std::setprecision(16) << "Computed integral value " << approx_integral
+                  << " does not match exact integral " << exact_integral << " for test problem " << std::endl
+                  << "∫_[-1,1]^n f(x[1],...,x[n]) * sinc(freq*x[n]) * ... * sinc(freq*x[n]) dx[1] ... dx[n]" << std::endl
```
mkstoyanov, line 118: This is very verbose, but OK. As a general note, when you debug this it is better to know which `f` failed rather than the general form of the integral. You can get the `f` from the dimension, so it's OK here, otherwise we would need another variable that is some form of a description.

Addons/testExoticQuadrature.hpp
```diff
@@ -0,0 +1,162 @@
...
+                  << "with inputs: " << std::endl << std::endl << std::left
```
mkstoyanov, line 119: Too much `std::endl` which is not the same as just a new line. The `std::endl` will also flush the I/O buffer, which adds extra overhead is not always needed. It is OK to have one `std::endl` at the end of the error message so the use will see the problem immediately, but adding two in a row or putting them in the middle of an I/O command is not good. Try:
```
<< "with inputs:\n\n" << std::left
```
and so on for the rest of the code.

Addons/testExoticQuadrature.hpp
```diff
@@ -0,0 +1,162 @@
...
+    }
+    return passed;
+}
+
+
+// Given a dimension, integrates the function f(x[1],...,x[dimension]) * sinc(freq * x[1]) * ... * sinc(freq * x[dimension]) over
+// [-1, 1]^dimension  using Exotic quadrature at a given depth and a specified shift. Then, compares this integral against a given exact
+// integral value.
+inline bool wrapSincTest(std::function<double(std::vector<double>)> f, int dimension, int depth, double freq, double shift,
+                         double exact_integral, double precision = 1e-10) {
```
mkstoyanov, line 93: Rename the `precision` to `tolerance`, otherwise things may get confusing with `precision` and `exactness`.

Addons/testExoticQuadrature.hpp
wwkong: Implemented by creating a more generic function called `wrapSincTest`.

Addons/tsgExoticQuadrature.hpp
wwkong: Changed.

Addons/testExoticQuadrature.hpp
wwkong: Noted.

Addons/testExoticQuadrature.hpp
wwkong: Changed the logic to use `const double*` instead of `std::vector<double>`. The definitions of `f1` and `f2` are slightly harder to read now.

Addons/testExoticQuadrature.hpp
wwkong: Removed the need to copy.

Addons/testExoticQuadrature.hpp
wwkong: Replaced all instances, except the last, of `std::endl` with `"\n"`.

Addons/testExoticQuadrature.hpp
wwkong: Done.

PR #609, opened 2021-08-31
Add symmetry template to getRoots() and function overloads for getExoticQuadrature()

wwkong:


Addons/tsgExoticQuadrature.hpp
```diff
@@ -97,23 +107,29 @@ inline double lagrange_eval(size_t idx, const std::vector<double> &roots, double
...
-//! The integrand is assumed to be of the form F(x) = f(x) * weight_fn(x) and the function x->weight_fn(x)+shift
-//! is nonnegative.
-inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double shift, std::function<double(double)> weight_fn,
-                                                    const int nref, const char* description) {
+//! The integral  I := ∫F(x)*weight_fn(x)dx, for a given level, is approximated as:
+//! I ≈ sum{f(roots[i])*(weight_fn_vals[i]+shift)*ref_weights[i]} - shift * sum{f(correction_points[i])*correction_weights[i]}.
+//!
+//! It is assumed that (weight_fn_vals[i]+shift) is nonnegative for every i.
+inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double shift, const std::vector<double> &weight_fn_vals,
+                                                    std::vector<double> &ref_weights, const std::vector<double> &ref_points,
```
mkstoyanov, line 118: Notice how everything `const` except `ref_weights` which gets modified, but it is not an output.

Quirks like this can waste a lot of time if someone is trying to figure out what's happening here. It may even be you and me couple of years from now.

Think of a way to solve this and we can discuss it in person.

wwkong: My first thought about this would be to make ``ref_weights`` a ``const`` input and just create a new array inside the function that stores the entries ``ref_weights[i] * (weight_fn_vals[i]  + shift)``, but this might not be the most efficient way. 

On the other hand, I'm not sure if I _can_ save on storage when all of the inputs are ``const`` lvalues at the moment.

mkstoyanov: Your first thought was better than the current solution, since the interface is simpler and more consistent and the extra copy is a very cheap price to pay.

But there is an even better way. Why not pass in the modified `ref_weights` directly. Think of how that would work.

wwkong: Gotcha. I could rename ``ref_weights`` to something like ``shifted_weights`` and have it already contain the values ``ref_weights[i] * (weight_fn_vals[i] + shift)``. Then, I would modify the overloaded function (which takes in a lambda) appropriately. 

I was slightly against going this direction because the user has to be very careful in making sure that this variable is generated properly, and there is no easy way to check to see if they made a mistake.

mkstoyanov: > I was slightly against going this direction because the user has to be very careful in making sure that this variable is generated properly, and there is no easy way to check to see if they made a mistake.

Your observation is spot on and this is a real issue. Think of how we can resolve that.

Addons/testExoticQuadrature.hpp
```diff
@@ -128,24 +133,31 @@ inline bool wrapSincTest(std::function<double(const double*)> f, int dimension,
 inline bool testAccuracy() {
     bool passed = true;
     int depth, dimension;
-    // 1D problem instances.
+    auto f1 = [](const double* x)->double{return std::exp(-(*x)*(*x));};
+    auto f2 = [](const double* x)->double{return std::exp(-(*x)*(*x)-(*(x+1))*(*(x+1)));};
```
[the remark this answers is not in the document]
wwkong: Done.

Addons/tsgExoticQuadrature.hpp
wwkong: Changed to use ``shifted_weights`` for now. Will make an external/internal API in another pull request based on our earlier in-person conversation.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -168,9 +181,28 @@ inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double sh
...
-                                                    std::function<double(double)> weight_fn) {
-    return getExoticQuadrature(n, shift, weight_fn, 50*n+1, "Exotic quadrature");
+
+// If only the weight function is provided, the reference points and weights are generated by the GL rule and the weight_fn values
+// are generated by applying the weight function to these reference points.
+inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double shift, std::function<double(double)> weight_fn,
+                                                    const int nref, const char* description, const bool is_symmetric = false) {
+    std::vector<double> shifted_weights, ref_points, weight_fn_vals(nref);
+    TasGrid::OneDimensionalNodes::getGaussLegendre(nref, shifted_weights, ref_points);
+    std::transform(ref_points.begin(), ref_points.end(), weight_fn_vals.begin(), [weight_fn](double x){return weight_fn(x);});
```
mkstoyanov, line 191: Do you need to wrap a lambda inside a lambda? Can you just pass in `weight_fn()` directly?

Addons/tsgExoticQuadrature.hpp
```diff
@@ -45,30 +45,40 @@ inline double poly_eval(const std::vector<double> &roots, double x) {
...
+//! \brief Generate the roots of all orthogonal polynomials up to degree n, where orthogonality is with respect to the
+// measure given by ref_weights. Specializes depending on if the underlying weight function is symmetric or not.
+template <bool is_symmetric>
 inline std::vector<std::vector<double>> getRoots(const int n, const std::vector<double> &ref_weights,
                                                  const std::vector<double> &ref_points) {
 
     // Compute the roots incrementally.
     assert(ref_points.size() == ref_weights.size());
     assert(ref_points.size() > 0);
+    if (not std::all_of(ref_weights.begin(), ref_weights.end(), [](double x){return x>=0.0;})) {
```
mkstoyanov, line 57: Right now, you are checking whether this is true in two places, here and in the external API call to `getExoticQuadrature()`. You only need one of those checks.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -45,30 +45,40 @@ inline double poly_eval(const std::vector<double> &roots, double x) {
...
         double sqr_offdiag_numr = 0.0;
         double sqr_offdiag_denm = 0.0;
         for (size_t j=0; j<ref_points.size(); j++) {
-            diag_numr += ref_points[j] * (poly_vals[j] * poly_vals[j]) * ref_weights[j];
-            diag_denm += (poly_vals[j] * poly_vals[j]) * ref_weights[j];
             sqr_offdiag_numr += (poly_vals[j] * poly_vals[j]) * ref_weights[j];
             sqr_offdiag_denm += (poly_m1_vals[j] * poly_m1_vals[j])  * ref_weights[j];
         }
-        diag.push_back(diag_numr / diag_denm);
         if (i >= 1) {
```
mkstoyanov, line 82: Do not execute the loop and compute `sqr_offdiag_numr` and `sqr_offdiag_denm` unless you will use them in the `if (i >= 1)` case. Basically, put the loop inside the if-statement.

Addons/tsgExoticQuadrature.hpp
wwkong: Good point. I changed it to pass ``weight_fn`` in directly.

Addons/tsgExoticQuadrature.hpp
wwkong: I decided to just keep the test in ``getExoticQuadrature`` since ``getRoots`` should probably remain internal (and hence, its errors should not be shown to the user)

Addons/tsgExoticQuadrature.hpp
wwkong: Done.

mkstoyanov:
Is this ready to merge? If so, you can remove the `[WIP]` in front. That indicates to me that I should merge unless I have more comments about the code.

PR #611, opened 2021-09-01
Add a Python CustomTabulated Class

wwkong:
Create Python code that wraps the C++ CustomTabulated class.

### DIFF
--- a/InterfacePython/TasmanianSG.py
+++ b/InterfacePython/TasmanianSG.py
@@ -2317,6 +2317,86 @@ class TasmanianSparseGrid:
 
         pAxisObject.imshow(ZZ, cmap=sCmap, extent=[fXmin, fXmax, fYmin, fYmax])
 
+class CustomTabulated:
+    def __init__(self, tasmanian_library=0):
+        '''
+        Constructor that creates an empty CustomTabulated instance.
+        
+        tasmanian_library: indicates the libtasmaniansparsegrids.so file
+        
+        if tasmanian_library is an int: look for the library in the
+                                        default locaiton
+                                        "libtasmaniansparsegrid.so"
+                                        or
+                                        the lirary install prefix given
+                                        to cmake
+
+        if tasmanian_library is a string: use it as path, for example:
+                TasmanianSparseGrid(
+                           tasmanian_library =
+                           "opt/Tasmanian/lib/libtasmaniansparsegrid.so"
+                          )
+
+        otherwise:  tasmanian_library must be an instance of ctypes.cdll
+                    this is useful when creating lots of instances of
+                    TasmanianSparseGrid in order to avoid having each
+                    instance load a sepatate copy of the common library
+
+        '''
+        if (isinstance(tasmanian_library, int)):
+            self.pLibTSG = cdll.LoadLibrary(TasmanianConfig.__path_libsparsegrid__)
+        elif (sys.version_info.major == 3):
+            if (isinstance(tasmanian_library, str)):
+                self.pLibTSG = cdll.LoadLibrary(tasmanian_library)
+            else:
+                self.pLibTSG = tasmanian_library
+        elif (isinstance(tasmanian_library, basestring)):
+            self.pLibTSG = cdll.LoadLibrary(tasmanian_library)
+        else:
+            self.pLibTSG = tasmanian_library
+
+        # ctypes requires that we manually specify the return types of functions
+        # in C this is done by header files, so this serves as a header
+        self.pLibTSG.tsgConstructCustomTabulated.restype = c_void_p;
+        self.pLibTSG.tsgGetNumLevelsCustomTabulated.restype = c_int;
+        self.pLibTSG.tsgGetNumPointsCustomTabulated.restype = c_int;
+        self.pLibTSG.tsgGetIExactCustomTabulated.restype = c_int;
+        self.pLibTSG.tsgGetQExactCustomTabulated.restype = c_int;
+        self.pLibTSG.tsgGetDescriptionCustomTabulated.restype = c_char_p;
+
+        self.pLibTSG.tsgDestructCustomTabulated.argtypes = [c_void_p];
+        self.pLibTSG.tsgGetNumLevelsCustomTabulated.argtypes = [c_void_p];
+        self.pLibTSG.tsgGetNumPointsCustomTabulated.argtypes = [c_void_p, c_int];
+        self.pLibTSG.tsgGetIExactCustomTabulated.argtypes = [c_void_p, c_int];
+        self.pLibTSG.tsgGetQExactCustomTabulated.argtypes = [c_void_p, c_int];
+        self.pLibTSG.tsgGetDescriptionCustomTabulated.argtypes = [c_void_p];
+
+        self.pCustomTabulated = self.pLibTSG.tsgConstructCustomTabulated()
+
+    def __del__(self):
+        '''
+        Destructor that calls the C++ destructor and releases all memory used by this instance of the class.
+        Make sure to call this to avoid memory leaks.
+        '''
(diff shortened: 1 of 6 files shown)


InterfacePython/TasmanianSG.py
```diff
@@ -50,6 +50,33 @@
 lsTsgLocalRules = ["localp", "semi-localp", "localp-zero", "localp-boundary"]
 lsTsgAccelTypes = ["none", "cpu-blas", "gpu-default", "gpu-cublas", "gpu-cuda", "gpu-rocblas", "gpu-hip", "gpu-magma"]
 
+def getTsgLib(tasmanian_library=0):
```
mkstoyanov, line 53: Putting the library inside the TasmanianSparseGrid class was an old thing and no longer needed, CMake and pip set the paths well enough so we don't have to do this trick of passing a custom path to the library.

Look at the more recent code inside TasmanianDreamState.py and TasmanianDreamSampler.py files and make the loaded library common to the module and not specific to the class.

InterfacePython/TasmanianSG.py
```diff
@@ -50,6 +50,33 @@
...
+            look for the library in the default location "libtasmaniansparsegrid.so" or the library install prefix given to CMake
+
+        if tasmanian_library is a string:
+            use it as a path; for example, TasmanianSparseGrid(tasmanian_library = "opt/Tasmanian/lib/libtasmaniansparsegrid.so")
+
+        otherwise:
+            tasmanian_library must be an instance of ctypes.cdll this is useful when creating lots of instances of TasmanianSparseGrid
+            in order to avoid having each instance load a sepatate copy of the common library
+    '''
+    if (isinstance(tasmanian_library, int)):
```
mkstoyanov, line 67: In Python, if you want to have a dummy parameter, do not use `int` use `None`.

SparseGrids/TasmanianSparseGridWrapC.cpp
```diff
@@ -478,6 +478,56 @@ void tsgGetGPUName(int gpu, int num_buffer, char *buffer, int *num_actual){
 
 void tsgDeleteInts(int *p){ delete[] p; }
 
+void* tsgConstructCustomTabulated(){ return (void*) new CustomTabulated(); }
+void tsgDestructCustomTabulated(void* ct){ delete ((CustomTabulated*) ct); }
+
+void tsgWriteCustomTabulated(void *ct, const char* filename, const bool use_binary_format){
```
mkstoyanov, line 484: Write only in ascii format here. The binary format is used when you want to save the whole grid.

SparseGrids/TasmanianSparseGridWrapC.cpp
```diff
@@ -478,6 +478,56 @@ void tsgGetGPUName(int gpu, int num_buffer, char *buffer, int *num_actual){
...
+int tsgGetNumLevelsCustomTabulated(void* ct){ return ((CustomTabulated*) ct)->getNumLevels(); }
+int tsgGetNumPointsCustomTabulated(void* ct, const int level){ return ((CustomTabulated*) ct)->getNumPoints(level); }
+int tsgGetIExactCustomTabulated(void* ct, const int level){ return ((CustomTabulated*) ct)->getIExact(level); }
+int tsgGetQExactCustomTabulated(void* ct, const int level){ return ((CustomTabulated*) ct)->getQExact(level); }
+const char* tsgGetDescriptionCustomTabulated(void* ct) { return ((CustomTabulated*) ct)->getDescription(); }
+
+void tsgGetWeightsNodesStaticCustomTabulated(void* ct, int level, double* w, double* x) {
+    int num_points = ((CustomTabulated*) ct)->getNumPoints(level);
+    std::vector<double> vec_w(num_points), vec_x(num_points);
+    ((CustomTabulated*) ct)->getWeightsNodes(level, vec_w, vec_x);
```
mkstoyanov, line 511: Update `CustomTabulated` to have an overload of `getWeightsNodes()` that directly write to arrays, no need to use vectors here and do an extra copy.

InterfacePython/TasmanianSG.py
```diff
@@ -2365,3 +2362,159 @@ def copyGrid(source, iOutputsBegin = 0, iOutputsEnd = -1):
...
+    filename: string indicating the location of the file.
+    tasmanian_library: see getTsgLib().
+
+    output: CustomTabulated
+    '''
+    ct = CustomTabulated(tasmanian_library)
+    ct.read(filename)
+    return ct
+
+def makeCustomTabulatedFromData(num_levels, num_nodes, precision, nodes, weights, description, tasmanian_library=0):
```
mkstoyanov, line 2487: This isn't directly related to the exotic quadrature, but it may be nice to have. It's OK to add.

InterfacePython/TasmanianSG.py
wwkong: Done.

InterfacePython/TasmanianSG.py
wwkong: Removed this line as part of the fix for the above issue.

SparseGrids/TasmanianSparseGridWrapC.cpp
wwkong: Done.

SparseGrids/TasmanianSparseGridWrapC.cpp
```diff
@@ -478,6 +478,56 @@ void tsgGetGPUName(int gpu, int num_buffer, char *buffer, int *num_actual){
 
 void tsgDeleteInts(int *p){ delete[] p; }
 
+void* tsgConstructCustomTabulated(){ return (void*) new CustomTabulated(); }
+void tsgDestructCustomTabulated(void* ct){ delete ((CustomTabulated*) ct); }
+
+void tsgWriteCustomTabulated(void *ct, const char* filename, const bool use_binary_format){
+    std::ofstream ofs(filename, std::ios::out);
```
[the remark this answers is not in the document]
wwkong: Done.

SparseGrids/TasmanianSparseGridWrapC.cpp
wwkong: Done.

InterfacePython/TasmanianSG.py
wwkong: Noted.

SparseGrids/tsgCoreOneDimensional.hpp
```diff
@@ -98,7 +98,14 @@ class CustomTabulated{
     int getQExact(int level) const{ checkLevel(level, "q-exactness"); return precision[level]; }
 
     //! \brief Get the points \b x and quadrature weights \b w associated with the rule at the \b level.
-    void getWeightsNodes(int level, std::vector<double> &w, std::vector<double> &x) const;
+    void getWeightsNodes(int level, std::vector<double> &w, std::vector<double> &x) const {
+        w.resize(getNumPoints(level));
+        x.resize(getNumPoints(level));
+        getWeightsNodes(level, w.data(), x.data());
```
mkstoyanov, line 104: No need to do that. Usually the vector method does this, but you can leave it as the `=` version that we had before. Basically, don't change this overload, just the other one.

wwkong: Done.

InterfacePython/sandbox.py
```diff
@@ -20,4 +20,22 @@
 # You can also use the file without cmake
 ###############################################################################
 
-print("Add code to this file to test, debug, or develop features")
+# print("Add code to this file to test, debug, or develop features")
+import Tasmanian
```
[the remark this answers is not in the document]
wwkong: Done.

PR #614, opened 2021-09-02
Add aliases for loadNeededPoints() in TasmanianSparseGrid (C++ & Python)

wwkong:


SparseGrids/TasmanianSparseGrid.hpp
```diff
@@ -813,6 +813,16 @@ class TasmanianSparseGrid{
      * Identical to loadNeededPoints() but does not throw if \b vals has an incorrect size (but will segfault).
      */
     void loadNeededPoints(const double *vals);
+    /*!
+     * \brief Alias of loadNeededPoints().
+     */
+    void loadNeededValues(std::vector<double> const &vals) {loadNeededPoints(vals);}
```
mkstoyanov, line 819: Add it to the list of functions in the documentation at the top of the TasmanianSparseGrid class.

wwkong: Done. I also removed the `getLoadedValues()` in the section **Load Model Values or Hierarchical Coefficients** as it doesn't seem like it belongs there.

mkstoyanov: I'm confused. 

`getLoadedValues()` allows you to get what you have loaded and it pairs with `loadNeededVlaues()`. Why do you think those don't belong together?

Also, `TasGrid::loadNeededPoints()` refers to a template in the Addons module that accepts a lambda expression to do the loading and it is not the method that is the member of the grid class. You can add an overload for the template, in this PR or another, but don't add it to the docs before it exists.

wwkong: Apologies. I thought that ``getLoadedValues()`` was: (i) redundant because it is referenced in the previous section and (ii) it does not "load" anything into the grid instance.

I have put it back and removed ``TasGrid::loadNeededValues()``.

mkstoyanov: The class has many-many methods and the purpose of the sections is to put those into categories so that the user can easily look at  a general high-level objective and see the closely related methods. Redundancy is good, if a method is relevant for two categories. The "relation" also goes both ways, if you load something it is natural to see if you can retrieve it back, which is what you want to do with the exotic quadrature.

mkstoyanov: Rebase to the latest master which fixes the ROCm error in the CI and we can properly test this.

I should not have let the ROCm testing stay broken for so long.

wwkong: Done.

InterfacePython/TasmanianSG.py
```diff
@@ -112,6 +112,7 @@
 pLibTSG.tsgGetInterpolationWeights.argtypes = [c_void_p, POINTER(c_double)]
 pLibTSG.tsgGetInterpolationWeightsStatic.argtypes = [c_void_p, POINTER(c_double), POINTER(c_double)]
 pLibTSG.tsgLoadNeededPoints.argtypes = [c_void_p, POINTER(c_double)]
+pLibTSG.tsgLoadNeededValues.argtypes = [c_void_p, POINTER(c_double)]
```
mkstoyanov, line 115: As I think about this, we don't need this one. In Python, we only need:
```
def loadNeededValues(self, llValues):
    self.loadNeededPoints(llValues)
```

Keep the C variant, it can be useful for other purposes.

wwkong: Done.

PR #617, opened 2021-09-07
[WIP] Add function overloads for getExoticQuadrature() + basic cleanup

wwkong:
Includes better documentation, more informative unit testing messages, and improved function signatures.


Addons/tsgExoticQuadrature.hpp
```diff
@@ -104,17 +116,24 @@ inline double lagrange_eval(size_t idx, const std::vector<double> &roots, double
...
-//!
-//! It is assumed that shifted_weights[i] is nonnegative and it is computed as shifted_weights[i] = ref_weights[i] *
-//! (weight_fn_vals[i] + shift) for every i, where ref_weights are the quadrature weights corresponding to ref_points.
-inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double shift, const std::vector<double> &weight_fn_vals,
-                                                    const std::vector<double> &shifted_weights, const std::vector<double> &ref_points,
-                                                    const int nref, const char* description, const bool is_symmetric = false) {
+
+/*!
+ * \internal
+ * \ingroup TasmanianAddons
```
mkstoyanov, line 122: Add a new Doxygen group `TasmanianAddonsExoticQuad` and use that group for all methods.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -178,28 +197,77 @@ inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double sh
...
 
-// If only the weight function is provided, the reference points are generated by the GL rule.
-
+/*
+ * \ingroup TasmanianAddons
+ * \brief Overloads of getExoticQuadrature() with some default values.
+ */
+inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double shift, TasGrid::TasmanianSparseGrid grid,
+                                                    const bool is_symmetric = false) {
+    return getExoticQuadrature(n, shift, grid, "Exotic quadrature", is_symmetric);
```
mkstoyanov, line 267: `ExoticQuadrature` is not a good name, the name appears in error messages and status reports from Tasmanian. It is important that the user provides a meaningful name and it is good to always require is as an input.

Addons/tsgExoticQuadrature.hpp
mkstoyanov: The mathematical explanation of exotic quadrature means should not be internal, it should be part of the description for the group.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -178,28 +197,77 @@ inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double sh
...
-// If only the weight function is provided, the reference points and weights are generated by the GL rule and the weight_fn values
-// are generated by applying the weight function to these reference points.
+/*
+ * \ingroup TasmanianAddons
+ * \brief Constructs a CustomTabulated object containing the exotic quadrature nodes and quadrature weights for a given level \b n,
+ * scalar \b shift, a reference TasmanianSparseGrid \b grid containing loaded values of the weight function, a string \b description,
+ * and an optional bool \b is_symmetric that should be set to true if the weight function is symmetric.
+ *
+ * It is assumed that grid has the following properties:
+ *     (i)   grid.getPoints() returns the quadrature reference points used to compute the entries of the Jacobi matrix.
```
mkstoyanov, line 207: (i) and (iii) combined means that you have a sparse grid interpolant/surrogate to the weight function. It is cleaner to say that you want an interpolant to the weight function, but we will only use the values at the points and the associated quadrature weights.

Also, the user wouldn't know what Jacobi matrix means. There are several Jacobi matrices in math, it is better to say that the point will be used to orthogonalize the polynomial basis needed to compute the optimal exotic quadrature.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -178,28 +197,77 @@ inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double sh
...
+ *     (ii)  grid.getQuadratureWeights() returns the quadrature reference weights corresponding to grid.getPoints().
+ *     (iii) grid.getLoadedValues() returns the values of the weight function at grid.getPoints().
+ */
+inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double shift, TasGrid::TasmanianSparseGrid grid,
```
mkstoyanov, line 211: Are you sure that you need to make a copy of the sparse grid as opposed to taking `const &`.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -178,28 +197,77 @@ inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double sh
...
+                                                    const char* description, const bool is_symmetric = false) {
+    std::vector<double> ref_points = grid.getPoints();
+    std::vector<double> shifted_weights = grid.getQuadratureWeights();
+    std::vector<double> weight_fn_vals = std::vector<double>(grid.getLoadedValues(), grid.getLoadedValues() + grid.getNumLoaded());
+    if (ref_points.size() == 0) {
+        throw std::invalid_argument("ERROR: grid needs to contain a nonzero number of nodes returned by grid.getPoints()!");
+    }
+    if (shifted_weights.size() != ref_points.size()) {
```
mkstoyanov, line 219: * start with the error checking and raise the appropriate exception before using `grid`, you may get an error when initializing the `ref_points` or `shifted_weights`.
* there's no need to check if the shifted weights and ref_points have the same number, the TasmanianSparseGrid class always returns the same number of points and weights (it's tested elsewhere).
* you need to test the folling
    * `grid.getNumLoaded() > 0` which means there are points loaded into the grid
    * `grid.getNumDimensions() == 1` since the weights must be single dimensional
    * `grid.getNumOutputs() == 1` since the output must be a scalar
    * positivity of the weights after the shift

Addons/tsgExoticQuadrature.hpp
```diff
@@ -178,28 +197,77 @@ inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double sh
...
+ * and an optional bool \b is_symmetric that should be set to true if the weight function is symmetric.
+ *
+ * It is assumed that grid has the following properties:
+ *     (i)   grid.getPoints() returns the quadrature reference points used to compute the entries of the Jacobi matrix.
+ *     (ii)  grid.getQuadratureWeights() returns the quadrature reference weights corresponding to grid.getPoints().
+ *     (iii) grid.getLoadedValues() returns the values of the weight function at grid.getPoints().
+ */
+inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double shift, TasGrid::TasmanianSparseGrid grid,
+                                                    const char* description, const bool is_symmetric = false) {
+    std::vector<double> ref_points = grid.getPoints();
```
mkstoyanov, line 213: This is not wrong, but for consistency we can use `getLoadedPoints()`

Addons/tsgExoticQuadrature.hpp
wwkong: > Add a new Doxygen group `TasmanianAddonsExoticQuad` and use that group for all methods.

Done.

> The mathematical explanation of exotic quadrature means should not be internal, it should be part of the description for the group.

Moved this to the newly created Doxygen group.

Addons/tsgExoticQuadrature.hpp
wwkong: I removed the simplifying overloads for now.

Addons/tsgExoticQuadrature.hpp
wwkong: I changed the descriptions according to your suggestions.

Addons/tsgExoticQuadrature.hpp
wwkong: Good point. I changed the function signature.

Addons/tsgExoticQuadrature.hpp
wwkong: Changed.

Addons/tsgExoticQuadrature.hpp
wwkong: Done.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -178,29 +193,62 @@ inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double sh
...
+ * and an optional bool \b is_symmetric that should be set to true if the weight function is symmetric.
+ *
+ * It is assumed that grid is a sparse grid interpolant/surrogate to the weight function. Only the nodes in grid.getLoadedPoints(),
+ * the quadrature weights in grid.getQuadratureWeights(), and the loaded weight function values in grid.getLoadedValues() are used
+ * in the computation of the exotic quadrature. Specifically, these vectors are used to orthogonalize the polynomial basis needed
+ * to compute the optimal exotic quadrature.
+ */
+inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double shift, const TasGrid::TasmanianSparseGrid &grid,
+                                                    const char* description, const bool is_symmetric = false) {
+    if (grid.getPoints().size() == 0) {
```
mkstoyanov, line 209: Important points:
* `getPoints()` may generate the "needed" points, i.e., the points that don't have loaded values and those cannot be used for the quadrature construction
* `getPoints()` and the other `getNeededPoints()` and `getLoadedPoints()` will do work and create a vector, but you only take the size of that vector

The correct method to use here is `grid.getNumLoaded()` which returns an `int` and it is an order 1 operation, i.e., just read a single number.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -178,29 +193,62 @@ inline TasGrid::CustomTabulated getExoticQuadrature(const int n, const double sh
...
 }
 
-// Basic function overloads.
+/*
+ * \ingroup TasmanianAddonsExoticQuad
+ * \brief Constructs a CustomTabulated object containing the exotic quadrature nodes and quadrature weights for a given level \b n,
+ * scalar \b shift, a reference TasmanianSparseGrid \b grid containing loaded values of the weight function, a string \b description,
+ * and an optional bool \b is_symmetric that should be set to true if the weight function is symmetric.
+ *
+ * It is assumed that grid is a sparse grid interpolant/surrogate to the weight function. Only the nodes in grid.getLoadedPoints(),
```
mkstoyanov, line 202: Instead of of `sparse grid` say `one dimensional`. When people hear `grid` they usually think of 2 or more dimensions and we want to emphasize that this is 1d.

Addons/tsgExoticQuadrature.hpp
wwkong: Changed.

Addons/tsgExoticQuadrature.hpp
wwkong: Done.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -92,7 +100,11 @@ inline std::vector<std::vector<double>> getRoots(const int n, const std::vector<
     return roots;
 }
 
-//! \brief Evaluates a Lagrange basis polynomial at a point \b x.
+/*!
+ * \internal
+ * \brief Evaluates a Lagrange basis polynomial at a point \b x.
```
mkstoyanov, line 105: Add the internal methods the Doxygen group. We'll improve the documentation later, but we can put things in the right spot now.

wwkong: Done.

PR #619, opened 2021-09-08
Add Python wrappers to the main C++ getExoticQuadrature() functions

wwkong:


Addons/tsgCExoticQuadrature.cpp
```diff
@@ -0,0 +1,55 @@
...
+#define __TASMANIAN_ADDONS_CEXOTICQUADRATURE_CPP
+
+#include "TasmanianAddons.hpp"
+
+using tsg_weight_fn = double(*)(double);
+extern "C"{
+    void* tsgCreateExoticQuadratureFromGrid(const int level, const double shift, void* grid_ptr,
+                                           const char* description, const bool is_symmetric) {
+        TasGrid::CustomTabulated *ct = new TasGrid::CustomTabulated;
+        TasGrid::TasmanianSparseGrid &grid = *reinterpret_cast<TasGrid::TasmanianSparseGrid*>(grid_ptr);
```
mkstoyanov, line 44: No need to declare a separate variable here, you can just put the `reinterpret_cast` statement in the function call below.

You can also rename `grid_ptr` to just `grid`.

InterfacePython/TasmanianAddons.py
```diff
@@ -402,3 +409,46 @@ def loadUnstructuredDataL2(points, model_data, tolerance, grid):
...
+    description:  string describing the Exotic quadrature instance.
+    is_symmetric: (optional) boolean that should be set to True if the weight function is symmetric.
+
+    output:       a Python CustomTabulated object.
+    '''
+    if not hasattr(ref_grid, "TasmanianSparseGridObject"):
+        raise TasmanianInputError("ref_grid", "ERROR: ref_grid must be an instance of TasmanianSparseGrid")
+    effective_description = bytes(description, encoding='utf8') if sys.version_info.major == 3 else description
+    ct = TasmanianSG.CustomTabulated()
+    ct.pCustomTabulated = pLibCTSG.tsgCreateExoticQuadratureFromGrid(c_int(level), c_double(shift), c_void_p(ref_grid.pGrid),
```
mkstoyanov, line 431: This is leaking memory.

Both `TasmanianSG.CustomTabulated()` and `pLibCTSG.tsgCreateExoticQuadratureFromGrid()` are calling `new` on C++ pointers, but `ct` has only one destructor and only one C++ `delete` will be issued.

Addons/tsgCExoticQuadrature.cpp
wwkong: Done.

InterfacePython/TasmanianAddons.py
wwkong: Changed the C functions to modify ``ct.pCustomTabulated`` directly rather than having them return a new object on the heap.

mkstoyanov:
You keep forgetting to remove the `[WIP]`

wwkong:
> You keep forgetting to remove the `[WIP]`

Should I remove it if I believe that I am done with my changes or only if all comments are resolved?

mkstoyanov:
> > You keep forgetting to remove the `[WIP]`
> 
> Should I remove it if I believe that I am done with my changes or only if all comments are resolved?

No! Think of it as `[WIP]` is your approval while the merge is my approval.

`[WIP]` comes from you and means a combination of the following:
* look at the PR to see where I'm headed, but I am not done yet
* I know there are problems still and I am working on those
* I don't intend to ever merge this one, but want it there for testing

Removing the `[WIP]` means that you are done with everything on your side and now it's my turn to read and agree/disagree. As a reviewer, I will not merge the PR until my concerns are fulfilled regardless of the `[WIP]` status.

PR #620, opened 2021-09-09
Add a variant of decompose() based on Golub's 1967 ALGOL code.

wwkong:
**Note**: some code has been added to `tasgridExternalTests.cpp` for verification purposes. We can remove it in the final pull request.

### DIFF
--- a/SparseGrids/tsgCoreOneDimensional.hpp
+++ b/SparseGrids/tsgCoreOneDimensional.hpp
@@ -205,7 +205,7 @@ namespace OneDimensionalMeta{
 namespace OneDimensionalNodes{
     // non-nested rules
     //! \brief Generate Gauss-Legendre weights \b w and points \b x for (input) number of points \b m.
-    void getGaussLegendre(int m, std::vector<double> &w, std::vector<double> &x, int version=1);
+    void getGaussLegendre(int m, std::vector<double> &w, std::vector<double> &x, int version=2);
     //! \brief Generate Chebyshev weights \b w and points \b x for (input) number of points \b m.
     void getChebyshev(int m, std::vector<double> &w, std::vector<double> &x);
     //! \brief Generate Gauss-Chebyshev type 1 weights \b w and points \b x for (input) number of points \b m.
@@ -213,11 +213,11 @@ namespace OneDimensionalNodes{
     //! \brief Generate Gauss-Chebyshev type 2 weights \b w and points \b x for (input) number of points \b m.
     void getGaussChebyshev2(int m, std::vector<double> &w, std::vector<double> &x);
     //! \brief Generate Gauss-Jacobi weights \b w and points \b x for (input) number of points \b m, using parameters \b alpha and \b beta.
-    void getGaussJacobi(int m, std::vector<double> &w, std::vector<double> &x, double alpha, double beta, int version=1);
+    void getGaussJacobi(int m, std::vector<double> &w, std::vector<double> &x, double alpha, double beta, int version=2);
     //! \brief Generate Gauss-Hermite weights \b w and points \b x for (input) number of points \b m, using parameters \b alpha
-    void getGaussHermite(int m, std::vector<double> &w, std::vector<double> &x, double alpha, int version=1);
+    void getGaussHermite(int m, std::vector<double> &w, std::vector<double> &x, double alpha, int version=2);
     //! \brief Generate Gauss-Laguerre weights \b w and points \b x for (input) number of points \b m, using parameters \b alpha
-    void getGaussLaguerre(int m, std::vector<double> &w, std::vector<double> &x, double alpha, int version=1);
+    void getGaussLaguerre(int m, std::vector<double> &w, std::vector<double> &x, double alpha, int version=2);
 
     // nested rules
     //! \brief Generate Clenshaw-Curtis \b nodes for the \b level.
--- a/SparseGrids/tsgCoreOneDimensional.hpp
+++ b/SparseGrids/tsgCoreOneDimensional.hpp
@@ -205,7 +205,7 @@ namespace OneDimensionalMeta{
 namespace OneDimensionalNodes{
     // non-nested rules
     //! \brief Generate Gauss-Legendre weights \b w and points \b x for (input) number of points \b m.
-    void getGaussLegendre(int m, std::vector<double> &w, std::vector<double> &x, int version=2);
+    void getGaussLegendre(int m, std::vector<double> &w, std::vector<double> &x);
     //! \brief Generate Chebyshev weights \b w and points \b x for (input) number of points \b m.
     void getChebyshev(int m, std::vector<double> &w, std::vector<double> &x);
     //! \brief Generate Gauss-Chebyshev type 1 weights \b w and points \b x for (input) number of points \b m.
@@ -213,11 +213,11 @@ namespace OneDimensionalNodes{
     //! \brief Generate Gauss-Chebyshev type 2 weights \b w and points \b x for (input) number of points \b m.
     void getGaussChebyshev2(int m, std::vector<double> &w, std::vector<double> &x);
     //! \brief Generate Gauss-Jacobi weights \b w and points \b x for (input) number of points \b m, using parameters \b alpha and \b beta.
-    void getGaussJacobi(int m, std::vector<double> &w, std::vector<double> &x, double alpha, double beta, int version=2);
+    void getGaussJacobi(int m, std::vector<double> &w, std::vector<double> &x, double alpha, double beta);
     //! \brief Generate Gauss-Hermite weights \b w and points \b x for (input) number of points \b m, using parameters \b alpha
-    void getGaussHermite(int m, std::vector<double> &w, std::vector<double> &x, double alpha, int version=2);
+    void getGaussHermite(int m, std::vector<double> &w, std::vector<double> &x, double alpha);
     //! \brief Generate Gauss-Laguerre weights \b w and points \b x for (input) number of points \b m, using parameters \b alpha
-    void getGaussLaguerre(int m, std::vector<double> &w, std::vector<double> &x, double alpha, int version=2);
+    void getGaussLaguerre(int m, std::vector<double> &w, std::vector<double> &x, double alpha);
 
     // nested rules
     //! \brief Generate Clenshaw-Curtis \b nodes for the \b level.
--- a/SparseGrids/tsgCoreOneDimensional.hpp
+++ b/SparseGrids/tsgCoreOneDimensional.hpp
@@ -205,7 +205,7 @@ namespace OneDimensionalMeta{
 namespace OneDimensionalNodes{
     // non-nested rules
     //! \brief Generate Gauss-Legendre weights \b w and points \b x for (input) number of points \b m.
-    void getGaussLegendre(int m, std::vector<double> &w, std::vector<double> &x);
+    void getGaussLegendre(int m, std::vector<double> &w, std::vector<double> &x, int version=1);
     //! \brief Generate Chebyshev weights \b w and points \b x for (input) number of points \b m.
     void getChebyshev(int m, std::vector<double> &w, std::vector<double> &x);
     //! \brief Generate Gauss-Chebyshev type 1 weights \b w and points \b x for (input) number of points \b m.
@@ -213,11 +213,11 @@ namespace OneDimensionalNodes{
     //! \brief Generate Gauss-Chebyshev type 2 weights \b w and points \b x for (input) number of points \b m.
     void getGaussChebyshev2(int m, std::vector<double> &w, std::vector<double> &x);
     //! \brief Generate Gauss-Jacobi weights \b w and points \b x for (input) number of points \b m, using parameters \b alpha and \b beta.
-    void getGaussJacobi(int m, std::vector<double> &w, std::vector<double> &x, double alpha, double beta);
+    void getGaussJacobi(int m, std::vector<double> &w, std::vector<double> &x, double alpha, double beta, int version=1);
     //! \brief Generate Gauss-Hermite weights \b w and points \b x for (input) number of points \b m, using parameters \b alpha
-    void getGaussHermite(int m, std::vector<double> &w, std::vector<double> &x, double alpha);
+    void getGaussHermite(int m, std::vector<double> &w, std::vector<double> &x, double alpha, int version=1);
     //! \brief Generate Gauss-Laguerre weights \b w and points \b x for (input) number of points \b m, using parameters \b alpha
-    void getGaussLaguerre(int m, std::vector<double> &w, std::vector<double> &x, double alpha);
+    void getGaussLaguerre(int m, std::vector<double> &w, std::vector<double> &x, double alpha, int version=1);
 
     // nested rules
     //! \brief Generate Clenshaw-Curtis \b nodes for the \b level.
--- a/SparseGrids/tsgLinearSolvers.cpp
+++ b/SparseGrids/tsgLinearSolvers.cpp
@@ -237,103 +237,110 @@ void TasmanianTridiagonalSolver::decompose(int n, std::vector<double> &d, std::v
     }
 }
 
-void decompose2(std::vector<double> &diag, std::vector<double> &off_diag, const double mu0, std::vector<double> &nodes,
-                std::vector<double> &weights) {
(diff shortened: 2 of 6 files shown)


SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -237,6 +237,111 @@ void TasmanianTridiagonalSolver::decompose(int n, std::vector<double> &d, std::v
...
+            continue;
+        }
+        // ALGOL COMMENT: Small off diagonal element means matrix can be split.
+        int k = m-1;
+        while (std::fabs(off_diag[k-1]) > eps) k--;
+        // ALGOL COMMENT: Find eigenvalues of lower 2-by-2 and select accelerating shift.
+        double b2 = off_diag[m-1] * off_diag[m-1];
+        double det = std::sqrt((diag[m-1] - diag[m]) * (diag[m-1] - diag[m]) + 4.0 * b2);
+        double aa = diag[m-1] + diag[m];
+        lambda2 = 0.5 * (aa >= 0 ? aa + det : aa - det);
```
mkstoyanov, line 286: As a note, this is using the `sign()` function:
```
lambda2 = 0.5 * ( aa + sign(aa) * det);
```

mkstoyanov:
This is not returning the same result as `decompose()`, run the tests locally on your machine and it should show up to test more easily.

You can run the tests with `ctest -V` or `./SparseGrid/gridtest global`

Look in the testing section of CMake to see the names for all the tests.

wwkong:
> This is not returning the same result as `decompose()`, run the tests locally on your machine and it should show up to test more easily.
> 
> You can run the tests with `ctest -V` or `./SparseGrid/gridtest global`
> 
> Look in the testing section of CMake to see the names for all the tests.

This should be fixed in the latest push. The issue was that the diagonal and off-diagonal entries had a small bug.

Everything looks okay except for the Gauss-Laguerre tests, which are off by a small tolerance (~1e-6 for nodes and ~1e-12 for weights).

SparseGrids/tsgCoreOneDimensional.hpp
```diff
@@ -205,19 +205,19 @@ namespace OneDimensionalMeta{
 namespace OneDimensionalNodes{
     // non-nested rules
     //! \brief Generate Gauss-Legendre weights \b w and points \b x for (input) number of points \b m.
-    void getGaussLegendre(int m, std::vector<double> &w, std::vector<double> &x);
+    void getGaussLegendre(int m, std::vector<double> &w, std::vector<double> &x, int version=1);
```
mkstoyanov, line 208: Adding the variable like that is a bad way to test things. If you want to test different algorithms, you can make a pre-processor macro or a `constexpr` internal variable, then you make one change in the code, recompile and have it work both ways all throughout.

This is just more typing and less efficient. It doesn't matter in the end, since we will not merge this and will have only one decompose in the end.

mkstoyanov: > Everything looks okay now, except for the Gauss-Laguerre tests, which are off by a small tolerance (~1e-6 for nodes and ~1e-12 for weights).

Gauss-Laguerre is one of the rules that works on an unbounded domain and it would make sense to be more sensitive. The question is whether your approach is more or less accurate than the original `decompose()`.

The original method has a tolerance parameter, can you check what happens if you make that one equal to machine `epsilon` or alternatively relax the tolerance in your approach to be `Maths::num_tol` which is about `1.E-12`.

Also, did you check Gauss-Chebyshev type 1? That weight is singular at the boundaries of the domain and I expect may be very sensitive to the eigenvalue method.

wwkong: > Adding the variable like that is a bad way to test things. If you want to test different algorithms, you can make a pre-processor macro or a `constexpr` internal variable, then you make one change in the code, recompile and have it work both ways all throughout.
> 
> This is just more typing and less efficient. It doesn't matter in the end, since we will not merge this and will have only one decompose in the end.

I'm not sure I understand. Wouldn't this force me to use only one of ``decompose()`` or ``decompose2()`` everywhere? Ideally, I would like to switch between the two based on a flag that I can change in the code, e.g., set the flag on and run ``getGaussLegendre()`` with ``decompose()``, then set the flag off and run ``getGaussLegendre()`` with ``decompose2()``, without recompiling in-between.

wwkong: > The original method has a tolerance parameter, can you check what happens if you make that one equal to machine `epsilon` or alternatively relax the tolerance in your approach to be `Maths::num_tol` which is about `1.E-12`.

Setting the original method's tolerance to machine `epsilon` and my method to `Maths::num_tol`, I saw a max error of `1.50067e-11` for the nodes and a max error of `5.77875e-09` for the weights at `n=1000` and `alpha=0.0`

> Also, did you check Gauss-Chebyshev type 1? That weight is singular at the boundaries of the domain and I expect may be very sensitive to the eigenvalue method.

The weights and nodes for all of the Gauss-Chebyshev variants are actually computed in closed form and don't need to invoke `decompose()`.

mkstoyanov: > I'm not sure I understand. Wouldn't this force me to use only one of `decompose()` or `decompose2()` everywhere? Ideally, I would like to switch between the two based on a flag that I can change in the code, e.g., set the flag on and run `getGaussLegendre()` with `decompose()`, then set the flag off and run `getGaussLegendre()` with `decompose2()`, without recompiling in-between.

My idea was to recompile and keep the intrusion to the minimum. All will be fixed in the end, but don't want to forget something.

I also don't want default values in the internal API, I got bitten too many times. If a method is internal, e.g., used only by the developers, then each input should be specified. Doesn't matter for intermediate testing, but it does for the final code that makes into `master`.

mkstoyanov: > The weights and nodes for all of the Gauss-Chebyshev variants are actually computed in closed form and don't need to invoke `decompose()`.

Gauss-Chebyshev type 1 is actually Gauss-Jacobi with `alpha = beta = -0.5`.

The closed form solution is perfect for testing the `decompose()` procedures, everything is best tested against a know closed form solution.

wwkong:
The accuracy of `decompose2()` for Gauss-Chebyshev Type I points looks good. On my machine, the maximum deviation between `getGaussJacobi()` with `alpha=beta=-0.5` and `getGaussChebyshev1()`  at `n=1000` is `9.88098e-15` for nodes and `1.63799e-12` for weights.

You can see the code in `tasgridExternalTests.cpp` in the latest push to check.

wwkong:
> Adding the variable like that is a bad way to test things. If you want to test different algorithms, you can make a pre-processor macro or a `constexpr` internal variable, then you make one change in the code, recompile and have it work both ways all throughout.

I've added a `constexpr` variable called `decompose_version` in `tsgLinearSolvers.hpp` that controls which version of `decompose` to use. The new one (Golub, 1967) is still called `decompose2`, but the old one is now called `decompose1`.

wwkong:
Using `decompose2()`, the tests for Gauss-Laguerre are the only ones that are failing, but only by a small amount. In particular, there are two failing instances with the following numbers:

**Instance 1**

Expected sum of weights:  `96.0 * 512.0 / 27.0` ~= `1820.444444444444`
Error of computed sum of weights: `4.09273e-12`
Tolerance for computed sum of weights: `1e-12`

Expected integral: `15360.0 * 3573248.0 / 243.0` ~= `225864564.9382716`
Error of computed integral: `1.37091e-06`
Tolerance for computed integral: `6e-7`

**Instance 2**

Expected sum of weights:  `96.0 * 512.0 / 27.0` ~= `1820.444444444444`
Error of computed sum of weights: `5.22959e-12`
Tolerance for computed sum of weights: `1e-12`

Based on the above, should I increase the tolerance for these tests or try make `decompose2()` more accurate?

mkstoyanov:
> 
> Based on the above, should I increase the tolerance for these tests or try make `decompose2()` more accurate?

That's a good question! I believe the tests look at absolute error, so looking at the magnitude of the answer and the corresponding error, we are within machine epsilon in relative error. But I don't like the fact that the error went up even by a bit, especially when we have exact analytic solution.

I'll have to take a closer look at this.

wwkong:
> > Based on the above, should I increase the tolerance for these tests or try make `decompose2()` more accurate?
> 
> That's a good question! I believe the tests look at absolute error, so looking at the magnitude of the answer and the corresponding error, we are within machine epsilon in relative error. But I don't like the fact that the error went up even by a bit, especially when we have exact analytic solution.
> 
> I'll have to take a closer look at this.

Maybe we can merge the current state of the PR (which uses the older version of `decompose()`), and do some adjustments in another PR?

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -237,6 +256,109 @@ void TasmanianTridiagonalSolver::decompose(int n, std::vector<double> &d, std::v
...
+
+    // SORT block from ALGOL code.
+    // ALGOL COMMENT: Arrange abscissas in ascending order.
+    // NOTE: the original code used an exchange sort, which is O(n^2).
+    std::vector<size_t> I(n);
+    for (size_t i=0; i<n; i++) I[i] = i;
+    std::sort(I.begin(), I.end(), [&nodes](size_t i, size_t j){return nodes[i] < nodes[j];});
+    for (size_t i=0; i<n; i++) {
+        if (i != I[i]) {
+            double tmp_node, tmp_weight;
```
mkstoyanov, line 343: Be careful when you read old code and old papers, especially in the 80s.

Right here, why declare the variables and only the initialize them?
Also this whole section seems it can be done with just a few swap operations suing `std::swap()`

wwkong: This sort block is actually my own implementation of sorting in-place (the original ALGOL code uses an O(n^2) exchange sort instead). Let me see if I can clean it up a little bit.

wwkong: Okay, I pushed a version that only uses only three swaps in the main while-loop. This one should be equivalent to the previous (longer) version.

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -164,7 +164,26 @@ std::vector<double> TasmanianTridiagonalSolver::getSymmetricEigenvalues(int n, s
     return result;
 }
 
-void TasmanianTridiagonalSolver::decompose(int n, std::vector<double> &d, std::vector<double> &e, std::vector<double> &z){
+void TasmanianTridiagonalSolver::decompose(std::vector<double> &diag, std::vector<double> &off_diag, const double mu0,
+                                           std::vector<double> &nodes, std::vector<double> &weights) {
+    switch(TasmanianTridiagonalSolver::decompose_version) {
+        case 1 :
+            weights.resize(diag.size());
```
mkstoyanov, line 171: This is messy.
* I think you want to use `weights = std::vector<double>(diag.size(), 0.0);` immediately followed by `weights[0] = mu0`.
    * resize will not initialize the vector to zero
    * the constructor uses `0.0` as initializer by default, but we explicitly need `0.0` here so it's better to say `0.0`
    * you don't want to be jumping between `weights` and `nodes` when setting variables

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -164,7 +164,26 @@ std::vector<double> TasmanianTridiagonalSolver::getSymmetricEigenvalues(int n, s
...
+            nodes.resize(diag.size());
+            nodes = diag;
```
mkstoyanov, line 173: Once you use the `=` operator, the nodes will be overwritten completely, no need to resize right before that.

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -237,6 +256,97 @@ void TasmanianTridiagonalSolver::decompose(int n, std::vector<double> &d, std::v
...
 }
 
+void TasmanianTridiagonalSolver::decompose2(std::vector<double> &diag, std::vector<double> &off_diag, const double mu0,
+                                            std::vector<double> &nodes, std::vector<double> &weights) {
+
+    // Ensure compatibility with the ALGOL implementation.
+    // NOTE: diag and off_diag are 1-indexed, while nodes and weights are 0-indexed after this step.
+    size_t n = diag.size();
+    if (off_diag.size() != n-1) {
+        throw std::invalid_argument("ERROR: off_diag.size() must be equal to diag.size()-1");
```
mkstoyanov, line 266: Use an `assert()`  in the internal API. Raising complex exceptions and doing the corresponding checks is wasteful, `assert()` is a cheap way for us to keep out code sane and the user can remove them at runtime. After all, if the rest of the code is correct, this will never happer.

* For external API, use informative error checking since the can mess up the inputs in all sorts of ways.
* For internal API, use `assert()` and only if you really need to use it.

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -237,6 +256,97 @@ void TasmanianTridiagonalSolver::decompose(int n, std::vector<double> &d, std::v
     }
 }
 
+void TasmanianTridiagonalSolver::decompose2(std::vector<double> &diag, std::vector<double> &off_diag, const double mu0,
+                                            std::vector<double> &nodes, std::vector<double> &weights) {
+
+    // Ensure compatibility with the ALGOL implementation.
+    // NOTE: diag and off_diag are 1-indexed, while nodes and weights are 0-indexed after this step.
```
mkstoyanov, line 263: This is beyond confusing, use C++ style of 0-indexing everywhere. It may be hard to read the Fortran code, but you can adjust the indexing.

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -237,6 +256,97 @@ void TasmanianTridiagonalSolver::decompose(int n, std::vector<double> &d, std::v
...
+    double lambda{norm}, lambda1{norm}, lambda2{norm}, rho{norm};
+
+    // INSPECT block from ALGOL code.
+    // ALGOL COMMENT: Look for convergence of lower diagonal element.
+    while (m > 0) {
+        if (std::fabs(off_diag[m-1]) <= eps) {
+            nodes[m-1] = diag[m];
+            weights[m-1] = mu0 * weights[m-1] * weights[m-1];
+            rho = lambda1 < lambda1 ? lambda1 : lambda2;
+            m = m-1;
```
mkstoyanov, line 295: Use `m--;`

It's Ok to blindly copy-paste code in the first iteration, but then fix the style to match the language that you're using.

SparseGrids/tsgLinearSolvers.cpp
wwkong: I'm not sure if I want to do this. The ALGOL code does both 1-indexing and 0-indexing in some places because it stores an additional variable in some arrays. 

For example, in `diag`, the array contains `n+1` elements, where the last `n` elements are the actually diagonal elements and the first element is an auxiliary one used in the QR computation, but which needs to be in position `0`. In C++, I would need to somehow store that element in "position" `-1`.

mkstoyanov: This is confusing. You can leave it for now, but I want to fix it eventually.

SparseGrids/tsgLinearSolvers.cpp
wwkong: Fixed.

SparseGrids/tsgLinearSolvers.cpp
wwkong: Fixed.

SparseGrids/tsgLinearSolvers.cpp
wwkong: Changed. Note that I had to add `cassert` to the list of includes.

SparseGrids/tsgLinearSolvers.cpp
wwkong: Done.

SparseGrids/tsgLinearSolvers.cpp
```diff
@@ -31,6 +31,7 @@
 #ifndef __TASMANIAN_LINEAR_SOLVERS_CPP
 #define __TASMANIAN_LINEAR_SOLVERS_CPP
 
+#include <cassert>
```
mkstoyanov, line 34: Put this line in `tsgEnumerate.hpp`, that is a top level header and includes all the C++ STL heades that we need.

wwkong: Done.

PR #621, opened 2021-09-09
Add Python functions that construct a sparse grid from a CustomTabulated object

wwkong:
**Note**: a small unit test has been added to ``sandbox.py``. We can remove it in the final pull request.


SparseGrids/TasmanianSparseGridWrapC.cpp
```diff
@@ -108,6 +108,16 @@ void tsgMakeFourierGrid(void *grid, int dimensions, int outputs, int depth, cons
...
     ((TasmanianSparseGrid*) grid)->makeFourierGrid(dimensions, outputs, depth, depth_type, anisotropic_weights, limit_levels);
 }
+void tsgMakeGridFromCustomTabulated(void *grid, int dimension, int outputs, int depth, const char *sType, void *custom_tabulated,
+                                    const int *anisotropic_weights, const int *limit_levels) {
+    TypeDepth depth_type = IO::getDepthTypeString(sType);
+    #ifndef NDEBUG
+    if (depth_type == type_none){ cerr << "WARNING: incorrect depth type: " << sType << ", defaulting to type_iptotal." << endl; }
+    #endif // NDEBUG
+    ((TasmanianSparseGrid*) grid)->makeGlobalGrid(dimension, outputs, depth, depth_type,
+                                                  std::move(*reinterpret_cast<TasGrid::CustomTabulated*>(custom_tabulated)),
```
mkstoyanov, line 118: If you move here, that would leave `custom_tabulated` in an undefined state. While C++ knows how to handle the problem, Python will try to issue a delete command on the pointer (when calling the destructor of CustomTabulated and then the C/C++ method).

The safest option is to use a copy.

InterfacePython/TasmanianSG.py
```diff
@@ -713,6 +714,63 @@ def makeFourierGrid(self, iDimension, iOutputs, iDepth, sType, liAnisotropicWeig
 
         pLibTSG.tsgMakeFourierGrid(self.pGrid, iDimension, iOutputs, iDepth, c_char_p(sType), pAnisoWeights, pLevelLimits)
 
+    def makeGridFromCustomTabulated(self, iDimension, iOutputs, iDepth, sType, iCustomTabulated, liAnisotropicWeights=[], liLevelLimits=[]):
```
mkstoyanov, line 717: * the name is inconsistent, call this `makeGlobalGridCustom()`, Global grid is a type in itself that exists in documentation and we should be explicit that this is a different way to make the grid as opposed to a new grid type
* no need to copy-paste the documentation, refer to the inputs of `makeGlobalGrid()` and outline the difference, that way if we have to update the docs, we won't have to update them in two places (or forget one place)
* procedurize the error checking, make new method `_testMakeGlobalGrid()` and add the error checking conditions for the common parameters in one place (I should have done that with the other grid methods)
* note the use of underscore in front of a method that indicates the method is not intended to be called by the user, similar to `private` or `protected` in C++
*

InterfacePython/TasmanianSG.py
```diff
@@ -2353,6 +2411,15 @@ def makeWaveletGrid(iDimension, iOutputs, iDepth, iOrder=1, liLevelLimits=[]):
     grid.makeWaveletGrid(iDimension, iOutputs, iDepth, iOrder, liLevelLimits)
     return grid
 
+def makeGridFromCustomTabulated(iDimension, iOutputs, iDepth, sType, iCustomTabulated, liAnisotropicWeights=[], liLevelLimits=[]):
```
mkstoyanov, line 2414: The naming convention is a bit old, but the first character in the name indicates the type:
```
iOutputs is an integer
sType is a string
pGrid is an object of some other type (e.g., sparse grid)
liLevelLimits is a list of integers, or a numpy array
llfX is a double-list or more specifically 2D-array of floating point numbers
pCustomRule is an object of type Tasmanian.CustomTabulated()
```

SparseGrids/TasmanianSparseGridWrapC.cpp
wwkong: Done.

SparseGrids/TasmanianSparseGridWrapC.cpp
```diff
@@ -108,6 +108,15 @@ void tsgMakeFourierGrid(void *grid, int dimensions, int outputs, int depth, cons
     if (depth_type == type_none){ depth_type = type_level; }
     ((TasmanianSparseGrid*) grid)->makeFourierGrid(dimensions, outputs, depth, depth_type, anisotropic_weights, limit_levels);
 }
+void tsgMakeGridFromCustomTabulated(void *grid, int dimension, int outputs, int depth, const char *sType, void *custom_tabulated,
+                                    const int *anisotropic_weights, const int *limit_levels) {
+    TypeDepth depth_type = IO::getDepthTypeString(sType);
+    #ifndef NDEBUG
+    if (depth_type == type_none){ cerr << "WARNING: incorrect depth type: " << sType << ", defaulting to type_iptotal." << endl; }
+    #endif // NDEBUG
+    TasGrid::CustomTabulated ct_copy = *reinterpret_cast<TasGrid::CustomTabulated*>(custom_tabulated);
```
mkstoyanov, line 117: You can inline the copy and skip the extra variable:
```
((TasmanianSparseGrid*) grid)->makeGlobalGrid(dimension, outputs, depth, depth_type, CustomTabulated(ct_copy), anisotropic_weights, limit_levels);
```
The rest of the code is good.

wwkong: Done.

InterfacePython/TasmanianSG.py
```diff
@@ -713,6 +714,34 @@ def makeFourierGrid(self, iDimension, iOutputs, iDepth, sType, liAnisotropicWeig
 
         pLibTSG.tsgMakeFourierGrid(self.pGrid, iDimension, iOutputs, iDepth, c_char_p(sType), pAnisoWeights, pLevelLimits)
 
+    def makeGlobalGridCustom(self, iDimension, iOutputs, iDepth, sType, iCustomTabulated, liAnisotropicWeights=[], liLevelLimits=[]):
```
mkstoyanov, line 717: This should read `pCustomTabulated` when you prefix with `i` it leads people to think it is an integer.

wwkong: Done.

PR #626, opened 2021-09-14
Bug fixes + Python unit tests for the CustomTabulated class.

wwkong:
Includes some changes to the exceptions thrown and file input types.

### DIFF
--- a/InterfacePython/testBasicIO.py
+++ b/InterfacePython/testBasicIO.py
@@ -431,17 +431,25 @@ class TestTasClass(unittest.TestCase):
             precision = np.array([2*(j+1)-1 for j in range(i)])
             nodes = [create_nodes(j, rng) for j in num_nodes]
             weights = [create_weights(j, rng) for j in num_nodes]
-            ctA = TasmanianSG.makeCustomTabulatedFromData(num_levels, num_nodes, precision, nodes, weights, description)
+            ctA = TasmanianSG.makeCustomTabulatedFromData(num_levels, num_nodes, precision, nodes, weights, description)            # .
             ctB = TasmanianSG.CustomTabulated()
             ctA.write("testSave")
             ctB.read("testSave")
-            ttc.compareCustomTabulated(ctA, ctB)
+            for i in range(num_levels):
+                read_weights, read_nodes = ctB.getWeightsNodes(i)
+                np.testing.assert_almost_equal(read_weights, weights[i])
+                np.testing.assert_almost_equal(read_nodes, nodes[i])
         # Read and write from a file.
         ctA = TasmanianSG.makeCustomTabulatedFromFile(tdata.sGaussPattersonTableFile)
         ctB = TasmanianSG.CustomTabulated()
         ctA.write("testSave")
         ctB.read("testSave")
         ttc.compareCustomTabulated(ctA, ctB)
+        for i in range(ctB.getNumLevels()):
+            read_weights, read_nodes = ctB.getWeightsNodes(i)
+            grid = TasmanianSG.makeGlobalGrid(1, 0, i, "level", "gauss-patterson")
+            np.testing.assert_almost_equal(read_weights, grid.getQuadratureWeights())
+            np.testing.assert_almost_equal(read_nodes, grid.getPoints().flatten())
         # Test an error message from wrong read.
         try:
             ctB.read("Test_If_Bogus_Filename_Produces_an_Error")
@@ -453,15 +461,13 @@ class TestTasClass(unittest.TestCase):
         '''
         Test makeGlobalGridCustom(), which creates a grid from a CustomTabulated instance.
         '''
+        gridA = TasmanianSG.makeGlobalGrid(1, 1, 3, "level", "custom-tabulated", [], 0.0, 0.0, tdata.sGaussPattersonTableFile)
         ct = TasmanianSG.makeCustomTabulatedFromFile(tdata.sGaussPattersonTableFile)
-        gridA1 = TasmanianSG.TasmanianSparseGrid()
-        gridA1.makeGlobalGrid(1, 1, 3, "level", "custom-tabulated", [], 0.0, 0.0, tdata.sGaussPattersonTableFile)
-        gridB1 = TasmanianSG.makeGlobalGridCustom(1, 1, 3, "level", ct)
-        ttc.compareGrids(gridA1, gridB1)
-        gridA1 = TasmanianSG.TasmanianSparseGrid()
-        gridA1.makeGlobalGrid(2, 1, 3, "level", "custom-tabulated", [], 0.0, 0.0, tdata.sGaussPattersonTableFile)
-        gridB1 = TasmanianSG.makeGlobalGridCustom(2, 1, 3, "level", ct)
-        ttc.compareGrids(gridA1, gridB1)
+        gridB = TasmanianSG.makeGlobalGridCustom(1, 1, 3, "level", ct)
+        ttc.compareGrids(gridA, gridB)
+        gridA = TasmanianSG.makeGlobalGrid(2, 1, 3, "level", "custom-tabulated", [], 0.0, 0.0, tdata.sGaussPattersonTableFile)
+        gridB = TasmanianSG.makeGlobalGridCustom(2, 1, 3, "level", ct)
+        ttc.compareGrids(gridA, gridB)
 
     def performIOTest(self):
         self.checkMetaIO()
--- a/InterfacePython/testBasicIO.py
+++ b/InterfacePython/testBasicIO.py
@@ -442,6 +442,12 @@ class TestTasClass(unittest.TestCase):
         ctA.write("testSave")
         ctB.read("testSave")
         ttc.compareCustomTabulated(ctA, ctB)
+        # Test an error message from wrong read.
+        try:
+            ctB.read("Test_If_Bogus_Filename_Produces_an_Error")
+        except TasmanianSG.TasmanianInputError as TSGError:
+            TSGError.bShowOnExit = False
+            self.assertEqual(TSGError.sVariable, "sFilename", "Reading a bogus file properly failed, but the error information is wrong.")
 
     def performIOTest(self):
         self.checkMetaIO()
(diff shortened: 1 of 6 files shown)


SparseGrids/tsgCoreOneDimensional.cpp
```diff
@@ -67,13 +67,13 @@ template<bool iomode> void CustomTabulated::read(std::istream &is){
         std::string T;
         char dummy;
         is >> T;
-        if (!(T.compare("description:") == 0)){ throw std::invalid_argument("ERROR: wrong file format of custom tables on line 1"); }
+        if (!(T.compare("description:") == 0)){ throw std::runtime_error("ERROR: wrong file format of custom tables on line 1"); }
```
mkstoyanov, line 70: Do not change these, someone may already be trying to capture those exceptions. Effectively, this changes the external API.

Catch two exceptions in the `tsgReadCustomTabulated()`, that's a better way to do it.

InterfacePython/testCommon.py
```diff
@@ -137,6 +143,26 @@ def compareGrids(self, gridA, gridB, bTestRuleNames = True):
...
+    def compareCustomTabulated(self, ctA, ctB):
+        '''
+        Compaes two CustomTabulated instances by checking nodes, weights, and other metadata.
+
+        The test passes if the two instances are mathematically identical.
+        '''
+        # Test metadata.
+        self.assertEqual(ctA.getDescription(), ctB.getDescription(), "error in getDescription()")
+        self.assertEqual(ctA.getNumLevels(), ctB.getNumLevels(), "error in getNumLevels()")
+        if (ctA.getNumLevels() == 0): # Empty CustomTabulated instance. Nothing else to check.
```
mkstoyanov, line 155: No need to `return`, if `getNumLevels()` is zero then the for-loop range is zero and nothing will happen.

InterfacePython/testCommon.py
```diff
@@ -39,6 +39,12 @@ def compareGrids(self, gridA, gridB, bTestRuleNames = True):
             return
 
         # load test points (canonical domain only), make sure to avoid grid points, e.g., 1/2, 1/4, etc.
+        if (gridA.getNumDimensions() == 1):
```
mkstoyanov, line 42: I don't see where you're using these. If the grid has no loaded points, then evaluate and the points should never be referenced.

InterfacePython/testBasicIO.py
```diff
@@ -410,6 +410,59 @@ def checkReadWriteMisc(self):
...
+        except TasmanianSG.TasmanianInputError as TSGError:
+            TSGError.bShowOnExit = False
+            self.assertEqual(TSGError.sVariable, "sFilename", "Reading a bogus file properly failed, but the error information is wrong.")
+
+    def checkGlobalGridCustom(self):
+        '''
+        Test makeGlobalGridCustom(), which creates a grid from a CustomTabulated instance.
+        '''
+        ct = TasmanianSG.makeCustomTabulatedFromFile(tdata.sGaussPattersonTableFile)
+        gridA1 = TasmanianSG.TasmanianSparseGrid()
```
mkstoyanov, line 457: Use a more logical order of operations:
* initialize gridA1, you can also use `TasmanianSG.makeGlobalGrid()` instead of making an empty grid and then overwritting
* initialize `ct`
* use `ct` to initialize `gridB1`

Right now you start with `ct` but you don't use it until later which breaks the flow when someone is reading the code.

InterfacePython/testBasicIO.py
```diff
@@ -410,6 +410,59 @@ def checkReadWriteMisc(self):
...
+        gridA1.makeGlobalGrid(1, 1, 3, "level", "custom-tabulated", [], 0.0, 0.0, tdata.sGaussPattersonTableFile)
+        gridB1 = TasmanianSG.makeGlobalGridCustom(1, 1, 3, "level", ct)
+        ttc.compareGrids(gridA1, gridB1)
+        gridA1 = TasmanianSG.TasmanianSparseGrid()
```
mkstoyanov, line 461: There is no need to reinitialize the grid to empty, calling `makeGlobalGrid()` will completely overwrite the existing grid.

InterfacePython/testBasicIO.py
```diff
@@ -410,6 +410,59 @@ def checkReadWriteMisc(self):
...
+            num_levels = i
+            num_nodes = np.array([3*(j+1) for j in range(i)])
+            precision = np.array([2*(j+1)-1 for j in range(i)])
+            nodes = [create_nodes(j, rng) for j in num_nodes]
+            weights = [create_weights(j, rng) for j in num_nodes]
+            ctA = TasmanianSG.makeCustomTabulatedFromData(num_levels, num_nodes, precision, nodes, weights, description)
+            ctB = TasmanianSG.CustomTabulated()
+            ctA.write("testSave")
+            ctB.read("testSave")
+            ttc.compareCustomTabulated(ctA, ctB)
```
mkstoyanov, line 438: This is effectively the same test as the one below, i.e., makes sure that the data that you save matches the data that you read. However, you don't explicitly test if the loaded points and weights match what you set in the random vectors.

The read/write test is OK, but you have to check the loaded points and weights.

InterfacePython/testBasicIO.py
```diff
@@ -410,6 +410,59 @@ def checkReadWriteMisc(self):
...
+        # Read and write from a file.
+        ctA = TasmanianSG.makeCustomTabulatedFromFile(tdata.sGaussPattersonTableFile)
```
mkstoyanov, line 440: Both `read()` and `makeCustomTabulatedFromFile()` use the same code-path to read the data. If one of them fails and say reads the wrong number of points, then both will fail in the exact same way.

Read from the Gauss-Patterson table, but then compare the points and weights against 1D sparse grid using the gauss-patterson rule, which will not read from a file but load from a hard-coded table.

SparseGrids/tsgCoreOneDimensional.cpp
wwkong: Done.

InterfacePython/testCommon.py
wwkong: Changed.

InterfacePython/testBasicIO.py
wwkong: Changed, and renamed `gridA1` (resp. `gridB1`) to `gridA` (resp. `gridB`) to be consistent with the rest of the code.

InterfacePython/testBasicIO.py
wwkong: Changed.

InterfacePython/testBasicIO.py
wwkong: Added code to test if the read `CustomTabulated` instance has the right loaded weights and nodes.

InterfacePython/testBasicIO.py
wwkong: Done.

InterfacePython/testCommon.py
wwkong: For 1D grids, using this function without the change will return an error because `mX1` to `mX5` are needed in `aBatchPoints` (line 61), but are unintialized.

mkstoyanov: `aBatchPoints` is used only if `getNumLoaded() > 0` on line `102`.

It is better to move the definition of `aBatchPoints` inside the if-statement on line `102`, rather than to defined unused variables.

wwkong: Changed. I moved the if-statement to before the first use of `mX1`, moved `aBatchPoints` to be in this if-block, and removed the previous 1D points.

InterfacePython/testCommon.py
```diff
@@ -86,20 +84,22 @@ def compareGrids(self, gridA, gridB, bTestRuleNames = True):
         pB = gridB.getQuadratureWeights()
         np.testing.assert_equal(pA, pB, "Quadrature not equal", True)
 
-        pA = gridA.getInterpolationWeights(mX1)
-        pB = gridB.getInterpolationWeights(mX1)
-        np.testing.assert_equal(pA, pB, "Interpolation test 1 not equal", True)
+        # evaluate (values have been loaded)
+        if (gridA.getNumLoaded() > 0):
```
mkstoyanov, line 88: I made a mistake here. I didn't realize that the `mX` points are used even without loaded points.

In the current set of tests, the `getInterpolationWeights()` methods test the construction of the Lagrange polynomials which doesn't require `getNumLoaded() > 0`. We should test the Lagrange polynomials even in the case when using CustomTabulated rules.

Defining a set of test points in the 1D case is the correct thing to do here. Your first attempt was the right one.

wwkong: Gotcha. I changed the code back to add the 1D nodes instead of moving around  `aBatchPoints` and changing the location of several if-blocks.

InterfacePython/testBasicIO.py
```diff
@@ -410,6 +410,62 @@ def checkReadWriteMisc(self):
...
+            precision = np.array([2*(j+1)-1 for j in range(i)])
+            nodes = [create_nodes(j) for j in num_nodes]
+            weights = [create_weights(j) for j in num_nodes]
+            ctA = TasmanianSG.makeCustomTabulatedFromData(num_levels, num_nodes, precision, nodes, weights, description)            # .
+            ctB = TasmanianSG.CustomTabulated()
+            ctA.write("testSave")
+            ctB.read("testSave")
+            for i in range(num_levels):
+                read_weights, read_nodes = ctB.getWeightsNodes(i)
+                np.testing.assert_almost_equal(read_weights, weights[i])
```
mkstoyanov, line 437: This is failing now.

Are you sure `np.linspace()` does what you expect? Can it really return only one point?

wwkong:
The problem in CI tests appears to be an incompatibility between `ctypes` and `numpy.ctypeslib`. In particular, the issue is in how they generate C arrays (`numpy.ctypeslib` fails to use the right integer type, e.g. `int32` and `int64`, sometimes). Since `ctypes` is the main interface, I will rewrite the code tomorrow to use it instead of the convenience functions in `numpy.ctypeslib`.

mkstoyanov:
Good job on finding the issue. Using the correct integer type is important, you can either switch all calls to `ctypes` and remove `numpy.ctypeslib` or change all the calls to use `int32` explicitly.

wwkong:
There seems to be some sort of docker error in the NVIDIA container, but aside from that, the PR should be ready to go now.

mkstoyanov:
> There seems to be some sort of docker error in the NVIDIA container, but aside from that, the PR should be ready to go now.

I think there is an issue on one of the machines. Rerunning the tests works, I'll tell the guy running the CI.

PR #627, opened 2021-09-15
Bug fixes + Python unit tests for the `createExoticQuadrature` functions.

wwkong:


mkstoyanov:
This looks good. But this requires the other PR first.

InterfacePython/testAddons.py
```diff
@@ -136,8 +137,45 @@ def checkUnstructuredL2(self):
...
+            np.testing.assert_almost_equal(compute_integral(grid, integrand), integral ** 2, 10,
+                                           "Computed integral does not match exact integral in test instance " + ct.getDescription())
+
     def performAddonTests(self):
         self.checkLoadNeeded()
         self.checkConstruction()
         self.checkBatchConstruct()
         self.checkUnstructuredL2()
+        # Tests that require LAPACK/BLAS
+        if (tdata.bHasBlas):
```
mkstoyanov, line 180: Why do we require BLAS? Is it because of `decompose()`?

Perhaps we can merge `decompose2()` first and have the exotic quadrature available with and without BLAS.

wwkong: Sounds good.

mkstoyanov:
Let's rebase and merge this one.

wwkong:
> Let's rebase and merge this one.

Done.

InterfacePython/testAddons.py
mkstoyanov: Do we still require BLAS for the exotic quadrature?

wwkong: Yes. I just submitted another PR (#632) that gives the option for exotic quadrature to use `decompose` instead of BLAS.

wwkong:
I removed the requirement that BLAS needs to be enabled to test exotic quadrature and rebased to master again.

This PR should be ready to merge.

InterfacePython/testAddons.py
```diff
@@ -136,8 +136,43 @@ def checkUnstructuredL2(self):
...
+        sinc1s0 = lambda x : np.sinc(x / np.pi)
+        sinc10s1 = lambda x : np.sinc(10 * (x - 1) / np.pi)
+        max_level = 25
+        nref = 101
+        def create_surrogate(lambda_fn):
+            grid = Tasmanian.makeGlobalGrid(1, 1, nref, "level", "gauss-legendre")
+            grid.loadNeededValues(np.array([lambda_fn(x) for x in grid.getNeededPoints()]))
+            return grid
+        def compute_integral(grid, integrand):
+            return np.sum([integrand(x) for x in grid.getPoints()] * grid.getQuadratureWeights())
```
mkstoyanov, line 151: As a general rule in Python, it is better to avoid entry-wise operations and use full arrays. For example, all the integrand functions accept vectors so you can avoid the for-loop:
```
return np.sum(integrand(grid.getPoints().squeeze()) * grid.getQuadratureWeights())
```

InterfacePython/testAddons.py
```diff
@@ -136,8 +136,43 @@ def checkUnstructuredL2(self):
...
+    def checkExoticQuadrature(self):
+        # Initialize some useful constants and utility functions.
+        integrand = lambda x : np.exp(-np.linalg.norm(x) ** 2)
+        sinc1s0 = lambda x : np.sinc(x / np.pi)
+        sinc10s1 = lambda x : np.sinc(10 * (x - 1) / np.pi)
+        max_level = 25
+        nref = 101
+        def create_surrogate(lambda_fn):
+            grid = Tasmanian.makeGlobalGrid(1, 1, nref, "level", "gauss-legendre")
+            grid.loadNeededValues(np.array([lambda_fn(x) for x in grid.getNeededPoints()]))
```
mkstoyanov, line 148: You can also write this without the for operation, see the comment below. You have to be more careful about the dimensions here.

InterfacePython/testAddons.py
```diff
@@ -136,8 +136,43 @@ def checkUnstructuredL2(self):
...
+            return grid
+        def compute_integral(grid, integrand):
+            return np.sum([integrand(x) for x in grid.getPoints()] * grid.getQuadratureWeights())
+        # Each pair below consists of a 1D integral value (first) and a CustomTabulated object (second) used to approximate this integral.
+        testPairs = [
+            [1.4321357541271255, Tasmanian.createExoticQuadratureFromFunction(max_level, 0.0, sinc1s0, nref, "Sinc1s0-shift0-Symm-byFn", True)],
```
mkstoyanov, line 154: Use scientific notation, which will align the first column of numbers and make it cleaner visually, e.g.,
```
[1.4321357541271255E-00, Tasmanian ...
[6.4062055930705356E-02, Tasmanian ...
```

InterfacePython/testAddons.py
wwkong: I modified this to
```
return np.sum(np.apply_along_axis(integrand, 1, grid.getPoints()) * grid.getQuadratureWeights())
```
because the `integrand` function has different behaviors when applied pointwise than to a vector of points (generated by `np.squeeze()`.

InterfacePython/testAddons.py
wwkong: Done.

InterfacePython/testAddons.py
wwkong: Done. I changed this to:
```
grid.loadNeededValues(np.apply_along_axis(lambda_fn, 1, grid.getNeededPoints()))
```
to be in line with the above comment.

PR #632, opened 2021-09-24
Allow `getExoticQuadrature()` to use `decompose()` instead of `LAPACK` + Bug fixes.

wwkong:
This PR also includes some minor fixes in the testing code.

### DIFF
--- a/Addons/tsgExoticQuadrature.hpp
+++ b/Addons/tsgExoticQuadrature.hpp
@@ -73,22 +73,22 @@ inline double lagrange_eval(size_t idx, const std::vector<double> &roots, double
 /*!
  * \internal
  * \ingroup TasmanianAddonsExoticQuad
- * \brief Generate the nodes and weights of a Gaussian quadrature (up to degree n) whose weight function is evaluated on \b
- * ref_nodes and its corresponding values are stored in \b ref_weights. Specializes depending on if the underlying weight
+ * \brief Generate the points and weights of a Gaussian quadrature (up to degree n) whose weight function is evaluated on \b
+ * ref_points and its corresponding values are stored in \b ref_weights. Specializes depending on if the underlying weight
  * function is symmetric or not.
  * \endinternal
  */
 template <bool is_symmetric>
-inline void getGaussNodesAndWeights(const int n, const std::vector<double> &ref_nodes, const std::vector<double> &ref_weights,
-                                    std::vector<std::vector<double>> &nodes_cache, std::vector<std::vector<double>> &weights_cache) {
+inline void getGaussNodesAndWeights(const int n, const std::vector<double> &ref_points, const std::vector<double> &ref_weights,
+                                    std::vector<std::vector<double>> &points_cache, std::vector<std::vector<double>> &weights_cache) {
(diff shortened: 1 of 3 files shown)


Addons/tsgExoticQuadrature.hpp
```diff
@@ -130,29 +151,32 @@ inline double lagrange_eval(size_t idx, const std::vector<double> &roots, double
...
-inline TasGrid::CustomTabulated getShiftedExoticQuadrature(const int n, const double shift,
+inline TasGrid::CustomTabulated getShiftedExoticQuadrature(const int n, const double shift, const std::vector<double> &weight_fn_vals,
                                                            const std::vector<double> &shifted_weights, const std::vector<double> &ref_points,
                                                            const char* description, const bool is_symmetric = false) {
-
-    // Create the set of points (roots) for the first term.
+    #ifndef Tasmanian_ENABLE_BLAS
+    if (gauss_quadrature_version == 1) {
+        #ifndef NDEBUG
+        std::cout << "WARNING: BLAS acceleration is not enabled!\nSwitching to an exotic quadrature algorithm that does not"
```
mkstoyanov, line 160: This is way too much info to dump on the user. You can use it for yourself, but will not merge it into master.

We should not be generating a warning for something that we support and should work just fine. What is even worse, we are saying "Success!" right after the warning. Why would the user care and what difference does it make.

Addons/tsgExoticQuadrature.hpp
```diff
@@ -130,29 +151,32 @@ inline double lagrange_eval(size_t idx, const std::vector<double> &roots, double
  * (weight_fn_vals[i] + shift) for every i.
  * \endinternal
  */
-inline TasGrid::CustomTabulated getShiftedExoticQuadrature(const int n, const double shift,
+inline TasGrid::CustomTabulated getShiftedExoticQuadrature(const int n, const double shift, const std::vector<double> &weight_fn_vals,
```
mkstoyanov, line 154: Why do we need `weight_fn_vals` here. Are those used outside of the `assert()` statement?

I thought that at this stage, we already have the weight function merged into the `shifted_weights`.

Addons/tsgExoticQuadrature.hpp
wwkong: Gotcha. I removed the messages and moved this into `getGaussNodesAndWeights()` since that is where it is needed.

Addons/tsgExoticQuadrature.hpp
wwkong: I looked over the code and it looks like we don't need it after all. I removed it in the latest push.

wwkong:
Some additional changes in the latest push:

- I no longer check for BLAS in `testAddons.cpp` for the exotic quadrature tests.
- I made a fix to the correction points in `tsgExoticQuadrature.hpp`. The original code always sets the center point to `0.0`, while the fix only does this if `is_symmetric` is `true`.

mkstoyanov (approved):
Is this ready to merge?

wwkong:
> Is this ready to merge?

Yes, it should be ready.

PR #633, opened 2021-09-29
CLI for creating Exotic quadratures 

wwkong:
An example of a command line call is:
```
./tasgrid -meq -depth 5 -shift 1.0 -weightfile sinc0.txt -description "Test Exoquad" -symmetric -outfile exoquad.ct -print
```
where `sinc0.txt` contains a weight function surrogate/intepolant that has been generated beforehand (see [here](https://github.com/ORNL/TASMANIAN/files/7260218/sinc0.txt) for an example).

### DIFF
--- a/SparseGrids/CMakeLists.txt
+++ b/SparseGrids/CMakeLists.txt
@@ -69,7 +69,8 @@ add_executable(Tasmanian_tasgrid tasgrid_main.cpp
                                  tasgridTestFunctions.hpp
                                  tasgridTestFunctions.cpp
                                  tasgridWrapper.hpp
-                                 tasgridWrapper.cpp)
+                                 tasgridWrapper.cpp
+                                 tsgExoticQuadrature.hpp)
 add_executable(Tasmanian_gridtest gridtest_main.cpp
                                   TasmanianSparseGrid.hpp
                                   tasgridCLICommon.hpp
--- a/SparseGrids/CMakeLists.txt
+++ b/SparseGrids/CMakeLists.txt
@@ -69,8 +69,7 @@ add_executable(Tasmanian_tasgrid tasgrid_main.cpp
                                  tasgridTestFunctions.hpp
                                  tasgridTestFunctions.cpp
                                  tasgridWrapper.hpp
-                                 tasgridWrapper.cpp
-                                 tsgExoticQuadrature.hpp)
+                                 tasgridWrapper.cpp)
 add_executable(Tasmanian_gridtest gridtest_main.cpp
                                   TasmanianSparseGrid.hpp
                                   tasgridCLICommon.hpp
@@ -112,6 +111,9 @@ endif()
 target_link_libraries(Tasmanian_libsparsegrid  Tasmanian_dependencies)
 
 target_include_directories(Tasmanian_libsparsegrid PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/>)
+# the following are needed for Exotic quadrature only
+target_include_directories(Tasmanian_libsparsegrid PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../Addons>)
+target_include_directories(Tasmanian_libsparsegrid PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../DREAM>)
 
 set_target_properties(Tasmanian_libsparsegrid PROPERTIES OUTPUT_NAME "tasmaniansparsegrid"
                                                          SOVERSION ${Tasmanian_VERSION_MAJOR}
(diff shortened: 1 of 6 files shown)


mkstoyanov:
* It doesn't help to add a link to a binary file, there is nothing that the anyone can get out of it. Even I can only say that this looks like a Tasmanian sparse grid file because I can read the first few characters. This is especially true when the user will never have to manually write such file, this is done using Tasmanian operations only.
* The output of the custom rule should always be in ascii format, that way the user can easily see the points and weights if they want and the `makeGlobalGrid()` method only works with an ascii custom tabulated file. The binary format for the custom tabulated was added so that the custom rule can be saved together with the rest of the sparse grid.
* is-symmetric is a flag, which means that you don't have to pass 0 or 1 from the command line, doing so is messy (what happens if the user passes 8 or a string or a floating point number). Assume that things are not symmetric (which is the more general case) and then have a flag that says `-symmetric` and that will switch to the symmetric mode.

SparseGrids/CMakeLists.txt
```diff
@@ -111,6 +111,9 @@ endif()
 target_link_libraries(Tasmanian_libsparsegrid  Tasmanian_dependencies)
 
 target_include_directories(Tasmanian_libsparsegrid PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/>)
+# the following are needed for Exotic quadrature only
+target_include_directories(Tasmanian_libsparsegrid PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../Addons>)
```
mkstoyanov, line 115: Note that this is technically not a proper way to use CMake. The correct thing to do is to move `tasgrid` into a separate folder that gets includes after the others. Thus, we can keep the hierarchy of sub-modules, e.g., Sparse Grid then DREAM then the Addons.

Don't do that now, we will do it after the release is finalized.

SparseGrids/tasgridWrapper.cpp
```diff
@@ -362,6 +379,13 @@ void TasgridWrapper::createQuadrature(){
         cerr << "ERROR: createQuadrature" << endl;
     }
 }
+void TasgridWrapper::createExoticQuadrature(){
+    TasGrid::TasmanianSparseGrid weight_surrogate;
+    weight_surrogate.read(weightfilename.c_str());
+    // Depth = 2 * level - 1.
+    int level = depth % 2 == 1 ? (depth + 1) / 2 : depth / 2 + 1;
```
mkstoyanov, line 386: Why is level different from depth?

wwkong:
> * It doesn't help to add a link to a binary file, there is nothing that the anyone can get out of it. Even I can only say that this looks like a Tasmanian sparse grid file because I can read the first few characters. This is especially true when the user will never have to manually write such file, this is done using Tasmanian operations only.

Changed the attached file to be in ASCII format.

> * The output of the custom rule should always be in ascii format, that way the user can easily see the points and weights if they want and the `makeGlobalGrid()` method only works with an ascii custom tabulated file. The binary format for the custom tabulated was added so that the custom rule can be saved together with the rest of the sparse grid.

Changed.

> * is-symmetric is a flag, which means that you don't have to pass 0 or 1 from the command line, doing so is messy (what happens if the user passes 8 or a string or a floating point number). Assume that things are not symmetric (which is the more general case) and then have a flag that says `-symmetric` and that will switch to the symmetric mode.

Changed to use `-symmetric` as an optional flag.

SparseGrids/tasgridWrapper.cpp
wwkong: From what I understand, depth refers to (roughly) the number of points that make up the grid. For the exotic quadrature function though, level is the number of Gauss-Legendre correction nodes or the generated weighted Gauss nodes.

For example, if we wanted an exotic quadrature with ~15 nodes, we would want to generate 8 Gauss-Legendre nodes and 8 weighted Gauss nodes (using `decompose()`) and account for the fact that `0.0` might be common between these nodes. Hence, the "level" given to `getExoticQuadrature()` would be 8 instead of the desired depth of 15.

mkstoyanov: Sparse Grid logic seldom goes by "number of points", we usually look at "level" or "polynomial order".
For the exotic quadrature, we should go by "level = depth".

An alternative could be that we add the criteria of polynomial order, i.e., find the level that matches the order but we can leave that off for now.

On a side note, even if you want to go by the number of points, there's no way to guarantee that some of the points for the shifted weight would not match the Gauss-Legendre points, so the formula may be wrong. If you want to specify the points exactly, you will have to change the logic of the construction and stop the find-roots process when you have enough total points. Let's not do that.

wwkong: Gotcha. I changed level to be the same as depth in the latest push.

PR #637, opened 2021-10-14
Small bugs

mkstoyanov:
Note two of the changes here:
* the MKL_SYCL_ROOT variable should not longer be needed, I figured I can read this right from the environment
* the MATLAB interface has a nasty use case where folders have spaces in the names
    * since the excutable could not be escaped with `" "` I opted to use escape sequences in the folder names
    * when you use `" "` in the filenames, this a file-not-found problem

### DIFF
--- a/Doxygen/Installation.md
+++ b/Doxygen/Installation.md
@@ -135,7 +135,7 @@ will be build and `BUILD_SHARED_LIBS` is always defined defaulting to `ON` follo
   -D CMAKE_CUDA_COMPILER:PATH          (specify the CUDA nvcc compiler)
   -D CMAKE_Fortran_COMPILER:PATH       (specify the Fortran compiler)
   -D Tasmanian_ROCM_ROOT:PATH          (specify the path to the ROCm installation)
-  -D Tasmanian_MKL_SYCL_ROOT:PATH      (specify the path to the oneMKL installation)
+  -D MKLROOT:PATH                      (specify the path to the oneMKL installation)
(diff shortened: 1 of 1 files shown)


Config/CMakeIncludes/FindTasmanianDpcpp.cmake
```diff
@@ -15,10 +15,10 @@ if (NOT Tasmanian_dpcpp_compiler AND NOT Tasmanian_sycl_compiler)
     message(FATAL_ERROR "Tasmanian_ENABLE_DPCPP requires that the CMAKE_CXX_COMPILER is set to the Intel dpcpp compiler or a compatible sycl compiler.")
 endif()
 
-if (NOT DEFINED Tasmanian_MKL_SYCL_ROOT AND NOT "$ENV{Tasmanian_MKL_SYCL_ROOT}" STREQUAL "")
-    set(Tasmanian_MKL_SYCL_ROOT "$ENV{Tasmanian_MKL_SYCL_ROOT}" CACHE PATH "The root folder for the Intel oneMKL installation")
+if (NOT "${MKLROOT}" STREQUAL "")
```
dalg24, line 18: I think you can write
```cmake
if (NOT MKLROOT)

mkstoyanov: > I think you can write
> 
> ```cmake
> if (NOT MKLROOT)
> ```

You're right. It doesn't work well with `$ENV{}` but it works when it is a regular variable.

PR #641, opened 2021-10-15
Various suggestions in the exotic quadrature documentation

wwkong:


Addons/tsgExoticQuadrature.hpp
```diff
@@ -307,20 +307,20 @@ inline void shiftReferenceWeights(std::vector<double> const &vals, double shift,
...
  * \param num_levels is the number of levels that should be computed
- * \param shift is the shift that needs to be added to the weights to make it positive
+ * \param shift is a real-valued scalar that needs to be added to the weight function to make it positive
  * \param grid has dimension 1 and at least one loaded point, should use as many points as possible for better accuracy
  *        (up to some point where numerical stability is lost)
  * \param description is a human readable string that can be used to identify the quadrature rule (could be empty)
- * \param is_symmetric indicates whether to assume that the weight is symmetric which allows for some stability improvements;
- *        symmetry is defined as "even" function on [-1, 1] which means that all odd power polynomials integrate to zero;
+ * \param is_symmetric indicates whether to assume that the weight is symmetric, which allows for some stability improvements;
+ *        symmetry is defined as being an "even" function on [-1, 1], which means that all odd power polynomials integrate to zero;
```
mkstoyanov, line 318: Expressions such as `being an` often over-encumber the language more than help streamline the point.

How about:
```
symmetric means "even" on [-1, 1] which leads to all odd power polynomials integrating to zero
```

Addons/tsgExoticQuadrature.hpp
```diff
@@ -329,7 +329,7 @@ inline void shiftReferenceWeights(std::vector<double> const &vals, double shift,
  * Note that the values of the weight function will not be interpolated between the reference points.
  *
  * Examples of symmetric weight functions are \f$ \sin(x) \f$, \f$ \sin(x) / x \f$, constant weight, and
- * the Gauss-Chebyshev weights. On the other hand, \f$ \cos(x) \f$ is not symmetric with regards
+ * the Gauss-Chebyshev weights. On the other hand, \f$ \cos(x) \f$ is not symmetric with respect
```
mkstoyanov, line 332: When we say `symmetric with respect to` it is something like the origin or the y-axis or some other point.

Howe about:
```
is not symmetric in the context of this algorithm
```

Addons/tsgExoticQuadrature.hpp
```diff
@@ -285,13 +285,13 @@ inline TasGrid::CustomTabulated getShiftedExoticQuadrature(const int n, const do
...
  * \internal
  * \ingroup TasmanianAddonsExoticQuad
- * \brief Multiplies the reference quadrature weights by the shifted values of the weights function.
+ * \brief Multiplies the reference quadrature weights by the shifted values of the weight function.
  *
  * \param vals are the values of the exotic weight function at the reference quadrature points
- * \param shift is the shift magnitude that ensures all \b vals[i] + \b shift are positive
- * \param ref_weights are the reference quadrature weights, will be overwritten with the result ref_weights[i] * (vals[i] + shift)
+ * \param shift is a real-valued scalar that ensures all \b vals[i] + \b shift are positive
+ * \param ref_weights is the reference quadrature weights, will be overwritten with the result ref_weights[i] * (vals[i] + shift)
```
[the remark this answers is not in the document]
mkstoyanov: Yes, we should be consistent. Change those too.

Addons/tsgExoticQuadrature.hpp
wwkong: Sounds good.

Changed.

PR #659, opened 2022-02-03
Add an optimization module

wwkong:
This pull request contains:

- A new module named `Optimization` with compatible `CMakeLists.txt`
- A base class called `TasOptimization::Solver` for which all optimization algorithms will inherit from
- A derived class called `TasOptimization::ParticleSwarmSolver` that implements the particle swarm algorithm
- An example of how to use the particle swarm solver to minimize the six-hump camel function in `Optimization/testOptimization.cpp`


Optimization/tsgSolver.hpp
```diff
@@ -0,0 +1,92 @@
...
+namespace TasOptimization {
+
+// A function that evaluates a single point.
+using ObjectiveFunction = std::function<double(const std::vector<double> &x)>;
+
+class Solver {
+  public:
+    Solver();
+    Solver(ObjectiveFunction input_fn, std::vector<double> &input_x);
+    ~Solver();
```
mkstoyanov, line 48: virtual ~Solver() = default;

CMakeLists.txt
```diff
@@ -133,10 +133,10 @@ configure_file("${CMAKE_CURRENT_SOURCE_DIR}/Tasgrid/tasgridLogs.in.hpp" "${CMAKE
 # After CXX, the interfaces Fortran, Python, MATLAB can come in any order
 # Finally, add the master target to link to all of the above
 ########################################################################
-add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/SparseGrids" "${CMAKE_CURRENT_BINARY_DIR}/SparseGrids")
-add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/DREAM"       "${CMAKE_CURRENT_BINARY_DIR}/DREAM")
-add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/Addons"      "${CMAKE_CURRENT_BINARY_DIR}/Addons")
-add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/Tasgrid/"    "${CMAKE_CURRENT_BINARY_DIR}/Tasgrid")
+add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/SparseGrids"  "${CMAKE_CURRENT_BINARY_DIR}/SparseGrids")
```
mkstoyanov, line 136: Remove the extra spaces here, no need to realign.

DREAM/Optimization/tsgOptimizationUtils.hpp
```diff
@@ -0,0 +1,63 @@
...
+ * ACCOMPLISH THE INTENDED RESULTS OR THAT THE SOFTWARE OR ITS USE WILL NOT RESULT IN INJURY OR DAMAGE. THE USER ASSUMES
+ * RESPONSIBILITY FOR ALL LIABILITIES, PENALTIES, FINES, CLAIMS, CAUSES OF ACTION, AND COSTS AND EXPENSES, CAUSED BY, RESULTING
+ * FROM OR ARISING OUT OF, IN WHOLE OR IN PART THE USE, STORAGE OR DISPOSAL OF THE SOFTWARE.
+ */
+
+#ifndef __TASMANIAN_OPTIM_ENUMERATES_HPP
+#define __TASMANIAN_OPTIM_ENUMERATES_HPP
+
+#include "TasmanianDREAM.hpp"
+#include <random>
```
mkstoyanov, line 38: DREAM already pulls in `<random>`,  you should be able to remove it.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,168 @@
...
+#ifndef __TASMANIAN_PARTICLE_SWARM_SOLVER_CPP
+#define __TASMANIAN_PARTICLE_SWARM_SOLVER_CPP
+
+#include "tsgParticleSwarm.hpp"
+
+namespace TasOptimization {
+
+ParticleSwarmState::ParticleSwarmState() : num_iterations(0), num_dimensions(0), num_particles(0) {}
+
+void ParticleSwarmState::setParticlePositions(std::vector<double> &pp) {
```
mkstoyanov, line 43: * make the input const
* the if-condition should be `pp.size() == num_dimensions * num_particles`
* add an overload that accepts rvalue reference

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,168 @@
...
+ */
+
+#ifndef __TASMANIAN_PARTICLE_SWARM_SOLVER_CPP
+#define __TASMANIAN_PARTICLE_SWARM_SOLVER_CPP
+
+#include "tsgParticleSwarm.hpp"
+
+namespace TasOptimization {
+
+ParticleSwarmState::ParticleSwarmState() : num_iterations(0), num_dimensions(0), num_particles(0) {}
```
mkstoyanov, line 41: Require that the user provides num-dimensions and num-particle in the constructor.

num_iteration is not part of the data state, we don't need that for now.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if (pp.size() % num_dimensions != 0) {
+        throw std::runtime_error("Size of particle positions (" + std::to_string(pp.size()) + ") is not a multiple of the " +
+                                 "number of dimensions (" + std::to_string(num_dimensions) + ")");
+    }
+    particle_positions = pp;
+}
+
+
+void ParticleSwarmState::setParticleVelocities(std::vector<double> &pv) {
+    if (pv.size() % num_dimensions != 0) {
```
mkstoyanov, line 53: If you let `num_dimensions` be zero, all of those checks will fail badly. Do not allow the user to create an object without specifying the number of dimensions.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+void ParticleSwarmState::addParticlesInsideBox(int box_num_particles, std::vector<double> &box_lower, std::vector<double> &box_upper,
+                                               std::function<double(void)> get_random01) {
+    if (box_lower.size() != box_upper.size())
+        throw std::runtime_error("The size of box lower bounds and box upper bounds do not match");
+    if (num_dimensions == 0) num_dimensions = box_lower.size();
+    num_particles += box_num_particles;
+
+    std::vector<double> box_particle_positions(num_dimensions * box_num_particles);
+    std::generate(box_particle_positions.begin(), box_particle_positions.end(), get_random01);
```
mkstoyanov, line 85: This is less efficient than if you call `get_random01` inside the nexted-for-loop below. If you can make your method cycle through all the data once, you should not be cycling multiple times.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,168 @@
...
+    for (int i=0; i<box_num_particles; i++) {
+        for (int j=0; j<num_dimensions; j++) {
+            box_particle_positions[i * num_dimensions + j] *= (box_upper[i] - box_lower[i]);
+            box_particle_positions[i * num_dimensions + j] -= box_lower[i];
+        }
+    }
+    std::copy(box_particle_positions.begin(), box_particle_positions.end(), std::back_inserter(particle_positions));
```
mkstoyanov, line 92: Why adding more particles, it's OK to keep the number fixed.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    std::vector<double> box_particle_velocities(num_dimensions * box_num_particles);
+    std::generate(box_particle_velocities.begin(), box_particle_velocities.end(), get_random01);
+    std::copy(box_particle_velocities.begin(), box_particle_velocities.end(), std::back_inserter(particle_velocities));
+}
+
+void ParticleSwarm(BatchedObjectiveFunction f, int max_iterations, TasDREAM::DreamDomain inside, ParticleSwarmState &state,
+                   double inertia_weight, double cognitive_coeff, double social_coeff, std::function<double(void)> get_random01) {
```
mkstoyanov, line 100: `get_random01` should always gave a default value, look at DREAM.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    // Initialize.
+    size_t num_dimensions = (size_t) state.getNumDimensions();
+    size_t num_particles = (size_t) state.getNumParticles();
+    std::vector<double> particle_positions = state.getParticlePositions();
```
mkstoyanov, line 105: Do not make a copy, this defeats the purpose of the state container.

Have protected methods that return references to the internal vectors and work with those by making `ParticleSwarm` into a `friend` method.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,168 @@
...
+                                           social_coeff * rg * (best_position[j] - particle_positions[idx]);
+            }
+            for (size_t j=0; j<num_dimensions; j++) {
+                particle_positions[i * num_dimensions + j] += particle_velocities[i * num_dimensions + j];
+            }
+        }
+        state.setParticlePositions(particle_positions);
+        state.setParticleVelocities(particle_velocities);
+        // Update the best positions.
+        f(particle_positions, fvals);
```
mkstoyanov, line 147: Here, `f()` may be called with points that don't satisfy `inside()`.

We should have two methods here, one that uses an `inside` function and one that is unconstrained. The one that uses `inside()` should make a second check here and call `f()` only for the points that satisfy the constraint.

DREAM/Optimization/tsgParticleSwarm.hpp
```diff
@@ -0,0 +1,76 @@
...
+#define __TASMANIAN_PARTICLE_SWARM_SOLVER_HPP
+
+#include "tsgOptimizationUtils.hpp"
+
+namespace TasOptimization {
+
+class ParticleSwarmState {
+  public:
+    ParticleSwarmState();
+    ~ParticleSwarmState() = default;
```
mkstoyanov, line 44: There is no need to say this, since this doesn't need to be virtual.

DREAM/Optimization/tsgParticleSwarm.hpp
```diff
@@ -0,0 +1,76 @@
...
+    void addParticlesInsideBox(int num_particles, std::vector<double> &lower, std::vector<double> &upper,
+                               std::function<double(void)> get_random01 = TasDREAM::tsgCoreUniform01);
+
+  private:
+    std::vector<double> particle_positions, particle_velocities, best_particle_positions, best_position;
+    int num_iterations, num_dimensions, num_particles;
+};
+
+void ParticleSwarm(BatchedObjectiveFunction f, int max_iterations, TasDREAM::DreamDomain inside, ParticleSwarmState &state,
+                   double inertia_weight = 0.5, double cognitive_coeff = 2, double social_coeff = 0.5,
```
mkstoyanov, line 72: Do not assign default values to the coefficients here.

Tasgrid/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ add_executable(Tasmanian_tasgrid tasgrid_main.cpp
                                  tasgridWrapper.cpp)
 
 target_link_libraries(Tasmanian_tasgrid Tasmanian_addons)
-
```
mkstoyanov, line 19: Problem with your editor, the editor should automatically delete those.

Optimization/tsgSolver.hpp
wwkong: Removed.

CMakeLists.txt
wwkong: Done.

DREAM/Optimization/tsgOptimizationUtils.hpp
wwkong: Changed.

DREAM/Optimization/tsgOptimizationUtils.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace TasOptimization {
+
+using ObjectiveFunction = std::function<void(const std::vector<double> &x, double &fval)>;
```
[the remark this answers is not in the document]
wwkong: Changed.

DREAM/Optimization/tsgOptimizationUtils.hpp
```diff
@@ -0,0 +1,63 @@
...
+using BatchedObjectiveFunction = std::function<void(const std::vector<double> &x_batch, std::vector<double> &fval_batch)>;
+
+inline BatchedObjectiveFunction makeBatchedFunction(int num_dimensions, ObjectiveFunction f) {
+    return [=](const std::vector<double> &x_values, std::vector<double> &fval_values)->void {
+        if (x_values.size() % num_dimensions != 0) {
```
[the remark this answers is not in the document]
wwkong: I'm not sure about including the check yet. It may be the case that the user makes an ``ObjectiveFunction`` using this function and passes this around to different solver functions (e.g. ``ParticleSwarm``) with checking that the dimensions align.

I did change the check, though.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Done. I also did the same with some of the other set functions to be consistent.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: I'm not sure about requiring ``num_particles`` in the constructor. Ideally, we want to have state objects that are always valid at any point they are observed, so creating states where ``num_particles`` > 0 and ``particle_positions`` is empty seems like a bad idea.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Fixed.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Fixed.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: See my comment about ``num_particles`` above.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Should we have to do this if default values are already given in the associated header file?

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Done.

DREAM/Optimization/tsgParticleSwarm.hpp
wwkong: Changed.

DREAM/Optimization/tsgParticleSwarm.hpp
wwkong: Changed.

Tasgrid/CMakeLists.txt
wwkong: I checked the file in the master branch, and these spaces also exist there.

DREAM/Optimization/tsgParticleSwarm.cpp
mkstoyanov: The header is enough, in fact it won't compile if you set defaults twice (even if it is the same default).

I didn't see the header.

Tasgrid/CMakeLists.txt
mkstoyanov: You're right, this should not have happened. It's OK to delete, those will cause an issue one way or another.

DREAM/Optimization/tsgParticleSwarm.cpp
mkstoyanov: I'm not worried about `fvals`, I'm worried about `particle_positions`. The objective shouldn't have to sanitize the input, if the user says "consider this constraint" then the constraint should be respected. If the objective relies on a model or a surrogate that has limited area of validity, our algorithm should respect that.

DREAM/Optimization/tsgParticleSwarm.hpp
```diff
@@ -0,0 +1,92 @@
...
+    inline int getNumIterations() const {return num_iterations;}
+    inline int getNumDimensions() const {return num_dimensions;}
+    inline int getNumParticles() const {return num_particles;}
+
+    void setParticlePositions(const std::vector<double> &pp);
+    void setParticlePositions(const std::vector<double> &&pp) {setParticlePositions(pp);};
+    void setParticleVelocities(const std::vector<double> &pv);
+    void setParticleVelocities(const std::vector<double> &&pv) {setParticleVelocities(pv);};
+    void setBestParticlePositions(const std::vector<double> &bpp);
+    void setBestParticlePositions(const std::vector<double> &&bpp) {setBestParticlePositions(bpp);}
```
mkstoyanov, line 59: * rvalue reference should not be const here
* this still copies on the call to `setBestParticlePositions();`, the idea of the rvalue is to move without copy, you need a second version of `setBestParticlePositions();` that uses `best_position = std::move(bp);`

DREAM/Optimization/tsgOptimizationUtils.hpp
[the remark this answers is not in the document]
mkstoyanov: If the user makes multiple objective functions, it's the user's responsibility to keep track which one goes where.

From our perspective, we want a function that has certain properties and we promise to always provide the correct input.

DREAM/Optimization/tsgParticleSwarm.cpp
mkstoyanov: You're right that we want "valid" objects, but a state with no particles or dimensions is not valid either. Certainly, we cannot use that state in the swarm algorithm.

The solution is the following:
* set dimensions and number of particles in the constructor
* initialize the storage space (the internal vector) to the correct size and fill it with zeros (which is default)
* add a flag that says "uninitalized", e.g., `bool initialized` member variable
* inside the swarm algorithm, we can check if the object has been initialized or not and raise an error
* in some cases, we can initialize automatically, e.g., if we have a sparse grid surrogate we automatically know the hypercube
* the constructor should also have an overload that accepts a vector of ready particles by rvalue-reference

DREAM/Optimization/tsgOptimizationUtils.hpp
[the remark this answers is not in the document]
wwkong: Fair enough. I removed the check.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Done. Note that the default constructor has been removed.

DREAM/Optimization/tsgParticleSwarm.hpp
wwkong: Fixed. The checks in these set functions now come from the helper function `checkVarSize()` in `tsgOptimizationUtils.hpp`.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+    std::vector<double> &particle_velocities = state.getParticleVelocitiesRef();
+    std::vector<double> &best_particle_positions = state.getBestParticlePositionsRef();
+    std::vector<double> &best_position = state.getBestPositionRef();
+    std::vector<double> prev_best_vals(num_particles), cur_best_vals(num_particles);
+    std::vector<bool> prev_is_inside(num_particles), cur_is_inside(num_particles);
+    double best_fval = std::numeric_limits<double>::max();
+    ObjectiveFunctionConstrained f_constrained = makeObjectiveFunctionConstrained(num_dimensions, f, inside);
+
+    // Create a lambda that updates the best particle and swarm positions based on function values and domain information.
+    auto update_best_positions = [&](const std::vector<double> prev_vals, const std::vector<double> cur_vals,
```
mkstoyanov, line 96: This copies needlessly, pass by const-reference as opposed to copy.

When using a Lambda, you can rely on the capture `[&]` and you don't have to accept inputs unless you're going to call the method with different inputs. Below, the inputs are always the same so might as well capture those too.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+                                     const std::vector<bool> prev_domain, const std::vector<bool> cur_domain)->void {
+        for (size_t i=0; i<num_particles; i++) {
+            bool is_smaller = (cur_vals[i] < prev_vals[i]) || !prev_domain[i];
+            if (cur_domain[i] && is_smaller) {
```
mkstoyanov, line 100: Use `and` as opposed to `&&`.

The `&&` comes from the older version of Tasmanian, I don't want to change the code that is already written and/or mix the two in other files, but for a new file and new module, we can use the more expressive `and` , `or`.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+                std::copy(particle_positions.begin() + i * num_dimensions, particle_positions.begin() + (i + 1) * num_dimensions,
```
mkstoyanov, line 101: `std::copy_n` is a cleaner way to do this, you don't need the middle iterator, e.g.,:
```
std::copy_n(particle_positions.begin() + i * num_dimensions, num_dimensions, best_particle_positions.begin() + i * num_dimensions);
```

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+                    std::copy(particle_positions.begin() + i * num_dimensions, particle_positions.begin() + (i + 1) * num_dimensions,
+                              best_position.begin());
+                    best_fval = cur_vals[i];
+                }
+            }
+        }
+    };
+
+    // Initialize the best particle and swarm function values, while updating the input state.
+    f_constrained(best_particle_positions, prev_best_vals, prev_is_inside);
```
mkstoyanov, line 113: This should not be needed, if the user has provided an initialization of `best_particle_position` then the user should also provide the values. Also, the values would be know if those come from an earlier call to the particle swam candicate.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+    f_constrained(particle_positions, cur_best_vals, cur_is_inside);
+    update_best_positions(prev_best_vals, cur_best_vals, prev_is_inside, cur_is_inside);
+    prev_best_vals = cur_best_vals;
```
mkstoyanov, line 116: This is a bit redundant:
* let's just have "cur_vals" and "best_vals" and we overwrite "best_vals" if the corresponding "cur_val" is better, otherwise "cur_best" and "prev_best" is messy, after all best is best.
* if you want to make an update like this (and I don't think you should from the comment above), but if you do, then use `std::swap()` to avoid the full copy.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // Main algorithm starts here.
+    while(state.getNumIterations() < max_iterations) {
+        for (size_t i=0; i<num_particles; i++) {
+            for (size_t j=0; j<num_dimensions; j++) {
+                double rp = get_random01();
+                double rg = get_random01();
+                int idx = i * num_dimensions + j;
+                particle_velocities[idx] = inertia_weight * particle_velocities[idx] +
+                                           cognitive_coeff * rp * (best_particle_positions[idx] - particle_positions[idx]) +
+                                           social_coeff * rg * (best_position[j] - particle_positions[idx]);
```
mkstoyanov, line 128: So you have `i` and `j` as `size_t` which is 64-bit, but when you multiply them and get a much larger number, you have `idx` as an `int`.

Instead of having double-index logic, note that only `best_position` uses a single index, so you can put all of this in a single loop:
```
for(size_t i=0; i<num_particles * num_dimensions; i++){
...
            social_coeff * rg * (best_position[i % num_dimensions] - particle_positions[i]);
...
}
```

You can also use the `Utils::size_mult()` template in tsgUtils.hpp that will convert both `num_particles` and `num_dimensions` to `size_t` and avoid annoying warnings.

On a side-note, are you using `-Wall -Wextra -Wshadow` for your internal testing. Sometimes those are annoying, but sometimes they help catch bugs.

DREAM/Optimization/tsgOptimizationUtils.hpp
```diff
@@ -0,0 +1,87 @@
...
+inline ObjectiveFunctionConstrained makeObjectiveFunctionConstrained(int ndim, ObjectiveFunction f, TasDREAM::DreamDomain inside) {
+    return [=](const std::vector<double> &x_batch, std::vector<double> &fval_batch, std::vector<bool> &inside_batch)->void {
+        std::vector<double> candidate(ndim);
+        std::vector<double> single_val(1);
+        for (size_t i=0; i<fval_batch.size(); i++) {
+            std::copy(x_batch.begin() + i * ndim, x_batch.begin() + (i + 1) * ndim, candidate.begin());
+            inside_batch[i] = inside(candidate);
+            fval_batch[i] = std::numeric_limits<double>::max();
+            if (inside_batch[i]) {
+                f(candidate, single_val);
```
mkstoyanov, line 78: The logic is correct, but this violates the desire for batch calls, since you're making calls on a point-by-point basis.

Do one sweep and resolve the `inside` call for all points. Then make a second sweep where you copy all good points into a single vector and make a single call to `f()`. Copy the result back to `fval_batch` in the correct order. This incurs an extra copy, but it is still better to make batch calls with one copy as opposed to point-by-point calls.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+#ifndef __TASMANIAN_PARTICLE_SWARM_SOLVER_CPP
+#define __TASMANIAN_PARTICLE_SWARM_SOLVER_CPP
+
+#include "tsgParticleSwarm.hpp"
+
+namespace TasOptimization {
+
+
+ParticleSwarmState::ParticleSwarmState(int ndim, int npart):
+        particle_positions(std::vector<double>(npart * ndim)), particle_velocities(std::vector<double>(npart * ndim)),
```
mkstoyanov, line 43: This is fine here, just note that you don't need to call `std::vector<double>` here, it is enough to say `particle_positions(npart * ndim)`.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+        best_particle_positions(std::vector<double>(npart * ndim)), best_position(std::vector<double>(ndim)),
+        num_dimensions(ndim), num_particles(npart), num_iterations(0), initialized(false) {};
+
+
+ParticleSwarmState::ParticleSwarmState(int ndim, std::vector<double> &&pp, std::vector<double> &&pv):
+        num_dimensions(ndim), num_iterations(0), initialized(true) {
```
mkstoyanov, line 49: This constructor does not initialize `num_particles`!

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+    assert(particle_positions.size() % ndim == 0);
+    int npart = particle_positions.size() / ndim;
+    particle_positions = std::move(pp);
+    particle_velocities = std::move(pv);
+    best_particle_positions = std::vector<double>(npart * ndim);
```
mkstoyanov, line 54: If you declare the `int` variables before the vectors in the class definition, then you can initialize all of the vectors once together with `num_dimensions`. This will avoid the double-initialization, once to an empty vector and then to the actual vector. This includes the moved vectors too.

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    // Initialize the best particle and swarm function values, while updating the input state.
+    f_constrained(best_particle_positions, prev_best_vals, prev_is_inside);
+    f_constrained(particle_positions, cur_best_vals, cur_is_inside);
+    update_best_positions(prev_best_vals, cur_best_vals, prev_is_inside, cur_is_inside);
+    prev_best_vals = cur_best_vals;
+    prev_is_inside = cur_is_inside;
+
+    // Main algorithm starts here.
+    while(state.getNumIterations() < max_iterations) {
```
mkstoyanov, line 120: You keep track of all iterations inside the `state` but `max_iterations` exits at the level of the algorithm. Let's take a minimalist approach here and count the iterations only for this current call to `ParticleSwarm()`.

The user can always count the iterations on their own and make judgement call on how to count which iteration. For example, initial iterations can be done with less social bias since we have very bad idea about the global minimum, later iterations can increase the bias as we begin to converge on the solution. Should the initial and later iterations count the same? Let the user figure it out.

DREAM/Optimization/tsgOptimizationUtils.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+#ifndef __TASMANIAN_OPTIM_ENUMERATES_HPP
+#define __TASMANIAN_OPTIM_ENUMERATES_HPP
+
+#include "TasmanianDREAM.hpp"
+#include <assert.h>
+
+namespace TasOptimization {
+
+inline void checkVarSize(std::string var_name, int var_size, int exp_size) {
```
mkstoyanov, line 42: I like this, clean and simple. Just add the method name to the information string, right now the user will see a mismatch but wouldn't know the function call that generated that. The variable name is less informative as the name in our implementation can be different than the name that the user has chosen, but the method name is the same.

DREAM/tasdreamExternalTests.cpp
```diff
@@ -565,6 +565,32 @@ bool DreamExternalTester::performTests(TypeDREAMTest test){
...
     cout << "Put here testing code and call this with ./dreamtest debug" << endl;
+
+    // Parameter setup.
+    int num_dimensions = 2;
+    int num_particles = 100;
+    int max_iterations = 50;
+    std::vector<double> lower = {-3, -2};
+    std::vector<double> upper = {3, 2};
+    TasOptimization::ObjectiveFunctionSingle shc_single = // six-hump camel function
+            [](const std::vector<double> x)->double {
```
mkstoyanov, line 576: Again creating a copy of `x`.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: It seems like you are suggesting that we only have a local `iteration` variable in `ParticleSwarm()`, but how would the information be sent back to the user? Should I override `state.num_iterations`? Should I return `iteration`?

DREAM/Optimization/tsgParticleSwarm.cpp
```diff
@@ -0,0 +1,146 @@
...
+    assert(particle_positions.size() % ndim == 0);
+    int npart = particle_positions.size() / ndim;
+    particle_positions = std::move(pp);
+    particle_velocities = std::move(pv);
+    best_particle_positions = std::vector<double>(npart * ndim);
+    best_position = std::vector<double>(ndim);
+};
+
+
+void ParticleSwarmState::addParticlesInsideBox(const std::vector<double> &box_lower, const std::vector<double> &box_upper,
```
mkstoyanov, line 59: The name should be `setParticlesInsideBox` or initialize particle or something similar. We are no longer "adding".

DREAM/Optimization/tsgParticleSwarm.cpp
mkstoyanov: ParticleSwarm doesn't have a convergence criteria, hence if the user asks for 10 iterations then we do 10 iterations.

When the user has a criteria (which would be another method), we can return the number of iterations either with `return num_iters` or as an output variable. If we make it part of the state (last resort), we can have something like `getNumLastIterations()`, but we don't keep some "global" history.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Fixed. This function no longer has any inputs.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Changed.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Changed to only iterating over `i`.

wwkong: For my testing: yes, my flags from CMake are `CMAKE_CXX_FLAGS="-O3 -Wall -Wextra -Wshadow"`.

DREAM/Optimization/tsgOptimizationUtils.hpp
wwkong: Changed.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Noted. I usually like to be more expressive here so that I don't need to check the header file for the variable types.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Fixed.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Changed. I also added a few checks here in case the user passes invalid particle positions and/or particle velocities.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Changed. I also removed `num_iterations` from the `ParticleSwarm` class definition.

DREAM/tasdreamExternalTests.cpp
wwkong: Fixed.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: Changed.

DREAM/Optimization/tsgParticleSwarm.cpp
wwkong: This part is a bit tricky because we are passing `inside()` as an input. 

More specifically, this is because a particular best particle value should be finite only if the corresponding point is inside the domain prescribed (which will depend on `inside()`), and there is no guarantee that the user will always pass in the same `inside()` function between calls.

DREAM/Optimization/tsgParticleSwarm.hpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+    void setBestPosition(std::vector<double> &&bp) {
+        checkVarSize("ParticleSwarmState::setBestPosition", "best position", bp.size(), num_dimensions);
+        best_position = std::move(bp);
+    }
+
+    void setNumDimensions(const int nd) {num_dimensions = nd;};
+    void setNumParticles(const int np) {num_particles = np;};
+
+    void setParticlesInsideBox(const std::vector<double> &box_lower, const std::vector<double> &box_upper,
```
mkstoyanov, line 94: Rename to initialize particles inside box.

DREAM/Optimization/tsgParticleSwarm.hpp
```diff
@@ -0,0 +1,118 @@
...
+  protected:
+    inline std::vector<double> &getParticlePositionsRef() {return particle_positions;}
+    inline std::vector<double> &getParticleVelocitiesRef() {return particle_velocities;}
+    inline std::vector<double> &getBestParticlePositionsRef() {return best_particle_positions;}
+    inline std::vector<double> &getBestPositionRef() {return best_position;}
+
+  private:
+    bool initialized;
+    int num_dimensions, num_particles;
+    std::vector<double> particle_positions, particle_velocities, best_particle_positions, best_position;
```
mkstoyanov, line 109: Values of the objective function should be kept inside the state.

The values are not part of the state, but they are cache data associated with the state.

We should also have a way to reset the values.

DREAM/Optimization/tsgParticleSwarm.cpp
mkstoyanov: We are updating the state to hold cached values for the `inside()`.

DREAM/Optimization/tsgParticleSwarm.hpp
wwkong: See comment below.

wwkong:
The latest push introduces the following major changes:

- `best_position` has been removed and the entries of this variable now appear at the end of `best_particle_positions`. For example, if `num_dimensions=2` and `num_particles=10`, then `best_particle_positions` is of size `22` and the last two positions contain the entries of the best position in the swarm.

- The following cache variables have been added:  `cache_particle_fvals`, `cache_best_particle_fvals`, `cache_particle_inside`, `cache_best_particle_inside`

- Split the `initialized` state boolean in `tsgParticleSwarm.hpp` into four state variables `positions_initialized`, `velocities_initialized`, `best_positions_initialized`, `cache_initialized`
    - The first two can be modified by `setParticlePositions()`, `setParticleVelocities()`, and `initializeParticlesInsideBox()`.
    - The second two can be modified by the new `clearBestParticles()` and `clearCache()` functions as well as the main `ParticleSwarm()` function.

- The `f_constrained()` lambda in `tsgParticleSwarm.cpp` is now formed explicitly inside the `ParticleSwarm()` function.

- Lines 146-154 in `tsgParticleSwarm.cpp` set up the cache variables and best position variables.

mkstoyanov (approved):

PR #662, opened 2022-03-07
Add the computation of derivatives/Jacobians for global grids 

wwkong:
# Summary
This pull request (PR) adds a new method `TasGrid::TasmanianSparseGrid::differentiate()` which can be invoked for global grids to compute the Jacobian matrix at a given evaluation point. 

# Jacobian matrix

Given an evaluation point `y` and function `f`, the Jacobian matrix `J` at `y` for input/dimension `k` and output `l` is of the form:
```
J[l][k] = sum of (f_at_x[i][l] * w[i][k]) over i
``` 
where `f_at_x[i][l]` is the l-th component of `f` evaluated at the i-th loaded point and `w[i][k]` is the derivative weight for the i-th loaded point and k-th input/dimension. Note that only `w` depends on `y`.

# Details

Some key functions/classes in this PR are:

- `TasGrid::GridGlobal::differentiate()`: The main logic for computing derivatives. This is called by the `TasGrid::TasmanianSparseGrid` instance if the instance manages a global grid.
- `TasGrid::GridGlobal::getDerivativeWeights()`: Computes the derivative weights `w`.
- `TasGrid::CacheLagrangeDerivative`: A new cache class that stores the derivatives of the Lagrange polynomials across different loaded points and inputs/dimensions. Inherits from the existing `TasGrid::CacheLagrange` class. Note that this class may reuse the computations already done in its parent class `TasGrid::CacheLagrange`.

Some simple examples are also given inside of `ExternalTester::debugTest()`.

### DIFF
--- a/SparseGrids/TasmanianSparseGrid.cpp
+++ b/SparseGrids/TasmanianSparseGrid.cpp
@@ -454,6 +454,14 @@ void TasmanianSparseGrid::integrate(double q[]) const{
     }
 }
 
+void TasmanianSparseGrid::differentiate(const double x[], double jacobian[]) const {
+    if (isGlobal()) {
+        base->differentiate(x, jacobian);
+    } else {
+        throw std::runtime_error("ERROR: in differentiate(), jacobians/gradients are not available for this type of grid");
+    }
+}
+
 void TasmanianSparseGrid::evaluate(const std::vector<double> &x, std::vector<double> &y) const{
     if (x.size() != (size_t) getNumDimensions()) throw std::runtime_error("ERROR: in evaluate() x must match getNumDimensions()");
     y.resize((size_t) getNumOutputs());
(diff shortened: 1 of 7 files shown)


SparseGrids/TasmanianSparseGrid.cpp
```diff
@@ -454,6 +454,14 @@ void TasmanianSparseGrid::integrate(double q[]) const{
     }
 }
 
+void TasmanianSparseGrid::differentiate(const double x[], double jacobian[]) const {
+    if (isGlobal()) {
+        dynamic_cast<GridGlobal*>(base.get())->differentiate(x, jacobian);
```
mkstoyanov, line 459: Check line 1980 in TasmanianSparseGrid.hpp, you can use:
```
get<GridGlobal>()->differentiate(x, jacobian);
```

wwkong: Changed.

mkstoyanov (approved):

mkstoyanov:
Also check if this works for `clenshaw-curtis0` which is different rule with implicit functions that are zero at the end-points of the domain. Doesn't have to be in this pull-request, but should be added.

Mark this as `ready to review` otherwise I cannot merge it.

wwkong:
> Also check if this works for `clenshaw-curtis0` which is different rule with implicit functions that are zero at the end-points of the domain. Doesn't have to be in this pull-request, but should be added.

I have some tests in another branch that includes the `clenshaw-curtis0` rule. It seems to work with this code.

> Mark this as `ready to review` otherwise I cannot merge it.

Done.

PR #667, opened 2022-03-24
Add tests for the differentiate method in global grids

wwkong:
Possible Extension?
---
Should tests be added for the `TasGrid::GridGlobal::getDerivativeWeights()` function? For example, like how `type_nodal_interpolation` tests `getInterpolationWeights()` and `type_internal_interpolation` tests `evaluate()`?


mkstoyanov:
Yes `TasGrid::GridGlobal::getDerivativeWeights()` should also be tested. Doesn't have to be in a single PR, this one is large already. As you see in the other tests, the test for nodal and internal interpolation use the same depth and tolerance, which makes things easy.

You should also adjust the depths for the Fourier tests or have a way to skip those. The Fourier tests use the same test logic as the global grid and are failing since you expect 4 options now (Fourier gives only the old way of 3 depths and tolerances).

wwkong:
> Yes `TasGrid::GridGlobal::getDerivativeWeights()` should also be tested. Doesn't have to be in a single PR, this one is large already. As you see in the other tests, the test for nodal and internal interpolation use the same depth and tolerance, which makes things easy.

Okay, I'll add these tests in another PR.

> You should also adjust the depths for the Fourier tests or have a way to skip those. The Fourier tests use the same test logic as the global grid and are failing since you expect 4 options now (Fourier gives only the old way of 3 depths and tolerances).

I've added an extra flag to `ExternalTester::testGlobalRule()` that specifies whether or not tests should be done on the `differentiate()` methods. For Fourier grids, I've set this flag to be `false`.

wwkong:
I just noticed that the checks on Windows (2019) and MacOS are failing for the differentiation tests (the computed relative error is too high). Do you have any clue about what might be causing the issue?

mkstoyanov:
> I just noticed that the checks on Windows (2019) and MacOS are failing for the differentiation tests (the computed relative error is too high). Do you have any clue about what might be causing the issue?

That's odd, if the error was `1.E-6` or something small, then we can blame rounding, but `0.5` is massive error. Something is significantly wrong and it does not manifest under `gcc`. I'll need to investigate more carefully, have you tried using `clang` under Linux?

wwkong:
> > I just noticed that the checks on Windows (2019) and MacOS are failing for the differentiation tests (the computed relative error is too high). Do you have any clue about what might be causing the issue?
> 
> That's odd, if the error was `1.E-6` or something small, then we can blame rounding, but `0.5` is massive error. Something is significantly wrong and it does not manifest under `gcc`. I'll need to investigate more carefully, have you tried using `clang` under Linux?

Yes. By default, I'm compiling with clang (/bin/clang++). I'll see if I can debug the issue on my Windows machine.

wwkong:
> > I just noticed that the checks on Windows (2019) and MacOS are failing for the differentiation tests (the computed relative error is too high). Do you have any clue about what might be causing the issue?
> 
> That's odd, if the error was `1.E-6` or something small, then we can blame rounding, but `0.5` is massive error. Something is significantly wrong and it does not manifest under `gcc`. I'll need to investigate more carefully, have you tried using `clang` under Linux?

Okay, I think I've fixed it. The issue was that the error vector `err` and norm vector `nrm` were not allocated enough space, and I was writing into indices that were out of their range.

SparseGrids/gridtestExternalTests.cpp
```diff
@@ -54,6 +54,30 @@ std::vector<double> genRandom(int num_samples, int num_dimensions = 1){
...
+    for (size_t i=0; i<unique_nodes.size(); i++) {
+        points[Utils::size_mult(i, num_dimensions)] = unique_nodes[i];
+    }
+    std::vector<double> err(unique_nodes.size(), 0.0), nrm(unique_nodes.size(), 0.0);
+    std::vector<double> r(num_outputs * num_dimensions), y(num_outputs * num_dimensions);
+    for (size_t k=0; k<unique_nodes.size(); k++) {
+        f->getDerivative(&(points[Utils::size_mult(k, num_dimensions)]), r.data());
+        grid.differentiate(&(points[Utils::size_mult(k, num_dimensions)]), y.data());
+        for (int i=0; i<num_outputs*num_dimensions; i++) {
+            nrm[k] = std::max(nrm[k], std::fabs(r[i]));
```
mkstoyanov, line 72: Instead of storing `err` and `nrm` into vectors, define them as local constants in the for-loop for each k. Then you can merge the second for-loop into this one:
```
double nrm = 0.0;
double err = 0.0;
for(int i=0; i<num_outputs*num_dimensions; i++) { ... }
rel_err = std::max(rel_err, std::fabs(nrm) < Maths::num_tol ? err : err / nrm);
```

SparseGrids/gridtestExternalTests.cpp
```diff
@@ -104,6 +128,23 @@ ExternalTester::ExternalTester(int in_num_mc) : num_mc(in_num_mc), verbose(false
     #ifdef Tasmanian_ENABLE_DPCPP
     test_queue.init_testing();
     #endif
+    // Hardcoded test types.
+    integr_only = {type_integration};
```
mkstoyanov, line 132: When long names differ by only two characters, it becomes very hard to read. Try naming those `quad_only` and `interp_only`

SparseGrids/gridtestExternalTests.cpp
```diff
@@ -215,101 +262,125 @@ TestResults ExternalTester::getError(const BaseFunction *f, TasGrid::TasmanianSp
...
-        loadValues(f, grid);
-
-        std::vector<double> err(num_outputs, 0.0); // absolute error
-        std::vector<double> nrm(num_outputs, 0.0); // norm, needed to compute relative error
-
-        std::vector<double> test_x = genRandom(num_mc, num_dimensions);
-        std::vector<double> result_tasm(num_mc * num_outputs);
-        std::vector<double> result_true(num_mc * num_outputs);
+    }else if (type == type_internal_interpolation or type == type_internal_differentiation){
+        if (type == type_internal_differentiation and (!grid.isGlobal() or grid.isSequence())) {
```
mkstoyanov, line 266: I don't think the if-statement is correct, are you missing `()` around `isGlobal() or isSequence()`

SparseGrids/gridtestExternalTests.cpp
```diff
@@ -215,101 +262,125 @@ TestResults ExternalTester::getError(const BaseFunction *f, TasGrid::TasmanianSp
...
+            R.error = 0.0;
+        } else {
+            // load needed points
+            loadValues(f, grid);
+
+            std::vector<double> test_x = genRandom(num_mc, num_dimensions);
+            int num_entries = (type == type_internal_interpolation) ? num_outputs : num_outputs * num_dimensions;
+            std::vector<double> result_tasm(num_mc * num_entries);
+            std::vector<double> result_true(num_mc * num_entries);
+            std::vector<double> err(num_entries, 0.0); // absolute error
```
mkstoyanov, line 277: Similar story to above, no need to store the errors and norms into a vector. The only time when this is needed is when you are inside a parallel region and threads need to write to different locations.

SparseGrids/gridtestExternalTests.cpp
```diff
@@ -215,101 +262,125 @@ TestResults ExternalTester::getError(const BaseFunction *f, TasGrid::TasmanianSp
...
     TasGrid::TasmanianSparseGrid grid;
     TestResults R;
-    int num_global_tests = (interpolation) ? 3 : 1;
-    TestType tests[3] = { type_integration, type_nodal_interpolation, type_internal_interpolation };
+    size_t num_global_tests = tests.size();
     TasGrid::TypeDepth type = (rule == rule_fourier ? TasGrid::type_level : TasGrid::type_iptotal);
     std::vector<double> x = genRandom(f->getNumInputs());
     if (rule == rule_fourier){ for(int i=0; i<f->getNumInputs(); i++) x[i] = 0.5*(x[i]+1.0); }    // map to canonical [0,1]^d
     bool bPass = true;
+    bool integration(false), interpolation(false), differentiation(false);
```
mkstoyanov, line 327: You have these granular conditions of different possibilities, but do you need this?

Every time you're checking these variables you are checking for `(interpolation or differentiation)` together, and even in that case you're only using the test to determine the number of outputs.

Why not have a new variable called `num_outputs`, set it to zero and then loop over `tests` if you see a test with internal interpolation/differentiation, then you set `num_outputs = f->getNumOutputs()` and simplify the code below.

SparseGrids/gridtestExternalTests.cpp
```diff
@@ -335,108 +406,117 @@ bool ExternalTester::performGlobalTest(TasGrid::TypeOneDRule rule) const{
...
-        const int depths1[3] = { 22, 22, 22 };
-        const double tols1[3] = { 1.E-12, 1.E-10, 1.E-10 };
-        const int depths2[3] = { 22, 22, 22 };
-        const double tols2[3] = { 1.E-12, 1.E-09, 1.E-09 };
-        if (testGlobalRule(&f21nx2, oned, 0, alpha, beta, true, depths1, tols1) && testGlobalRule(&f21cos, oned, 0, alpha, beta, true, depths2, tols2)){
+        const int depths1[4] = { 22, 22, 22, 22 };
+        const double tols1[4] = { 1.E-12, 1.E-10, 1.E-10, 1.E-09 };
+        const int depths2[4] = { 22, 22, 22, 22 };
+        const double tols2[4] = { 1.E-12, 1.E-09, 1.E-09, 1.E-08 };
+        if (testGlobalRule(&f21nx2, oned, 0, alpha, beta, all_test_types, depths1, tols1) &&
```
mkstoyanov, line 434: It's Ok to break the row, but change `&&` to `and`.

SparseGrids/gridtestExternalTests.cpp
wwkong: Changed.

SparseGrids/gridtestExternalTests.cpp
wwkong: Changed. I also shorted the other test case names.

SparseGrids/gridtestExternalTests.cpp
wwkong: Fixed.

SparseGrids/gridtestExternalTests.cpp
wwkong: Changed. Note that this logic for the error vectors existed before this PR.

SparseGrids/gridtestExternalTests.cpp
wwkong: Changed. I made a new variable called `num_fn_outputs` (to differentiate it from the grid's `num_outputs`) to store this information.

SparseGrids/gridtestExternalTests.cpp
wwkong: Changed.

PR #671, opened 2022-04-05
Add implementation and tests for differentiation of sequence grids

wwkong:


SparseGrids/tsgGridSequence.cpp
```diff
@@ -859,6 +931,52 @@ void GridSequence::applyTransformationTransposed(double weights[]) const{
     }
 }
 
+void GridSequence::applyDifferentiationTransformation(double weights[], const int num_dimensions) const{
```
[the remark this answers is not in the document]
wwkong: That's a good point. I've merged the two transforms using templates and do not pass `num_dimensions` as an input in the latest push.

SparseGrids/tsgGridSequence.hpp
```diff
@@ -131,11 +133,44 @@ class GridSequence : public BaseCanonicalGrid{
         return cache;
     }
 
+    // The second argument should be the result of calling TasGrid::GridSequence::cacheBasisValues(x).
+    template<typename T>
+    std::vector<std::vector<T>> cacheBasisDerivatives(const T x[], std::vector<std::vector<T>> &value_cache) const {
```
mkstoyanov, line 138: Many loops, checks, branching, special cases, division by zero problems, the need to pass in `value_cache` ...

Years ago, when I was studying computer science, my teachers were telling me that I should spend 15% of my time coding and 85% thinking about what I am about to code.

How about we replace this template by the following:
```
    template<typename T>
    std::vector<std::vector<T>> cacheBasisDerivatives(const T x[]) const {
        std::vector<std::vector<T>> cache(num_dimensions);
        for(int j=0; j<num_dimensions; j++){
            cache[j].resize(max_levels[j] + 1);
            T this_x = x[j];
            T b = 1.0;
            T s = 1.0;
            cache[j][0] = 0.0;
            if (max_levels[j] > 0) cache[j][1] = 1.0 / coeff[1];
            for(int i=2; i <= max_levels[j]; i++){
                s *= (this_x - nodes[i-1]);
                b *= (this_x - nodes[i-2]);
                s += b;
                cache[j][i] = s / coeff[i];
            }
        }
        return cache;
    }
```

wwkong: Looks good to me. I didn't immediately see that the product rule could be done recursively here.

mkstoyanov: The Newton polynomials are inherently recursive, which will manifest in all aspects of the problem even when it is not immediately obvious. The recursion makes a big difference in speeding things up.

SparseGrids/tsgGridSequence.cpp
```diff
@@ -526,6 +559,45 @@ void GridSequence::integrate(double q[], double *conformal_correction) const{
     }
 }
 
+void GridSequence::differentiate(const double x[], double jacobian[]) const {
```
mkstoyanov, line 562: You start with the idea to multiply everything together and then divide out what you don't need, which leads you down a rabbit hole of checks and if-statements that must be put in place to avoid a division by zero. Effectively, you're doing more work than needed which leads you to need even more work.

How about multiplying only that which needs to be multiplied:
```
void GridSequence::differentiate(const double x[], double jacobian[]) const {
    std::vector<std::vector<double>> value_cache = cacheBasisValues<double>(x);
    std::vector<std::vector<double>> derivative_cache = cacheBasisDerivatives<double>(x);
    std::vector<double> diff_values(num_dimensions);
    std::fill_n(jacobian, num_outputs * num_dimensions, 0.0);
    int n = points.getNumIndexes();
    for(int i=0; i<n; i++) {
        const int* p = points.getIndex(i);
        const double *s = surpluses.getStrip(i);

        diff_values[0] = derivative_cache[0][p[0]];
        for(int j=1; j<num_dimensions; j++) diff_values[j] = value_cache[0][p[0]];

        for(int k=1; k<num_dimensions; k++){
            for(int j=0; j<k; j++) diff_values[j] *= value_cache[k][p[k]];
            diff_values[k] *= derivative_cache[k][p[k]];
            for(int j=k+1; j<num_dimensions; j++) diff_values[j] *= value_cache[k][p[k]];
        }

        for(int k=0; k<num_outputs; k++)
            for(int j=0; j<num_dimensions; j++)
                jacobian[k * num_dimensions + j] += s[k] * diff_values[j];
    }
}
```

SparseGrids/tsgGridSequence.cpp
```diff
@@ -171,7 +171,40 @@ void GridSequence::getInterpolationWeights(const double x[], double *weights) co
             weights[i] *= cache[j][p[j]];
         }
     }
-    applyTransformationTransposed(weights);
+    applyTransformationTransposed<0>(weights);
+}
+
+void GridSequence::getDifferentiationWeights(const double x[], double weights[]) const {
```
mkstoyanov, line 177: This should be fixed too.

SparseGrids/tsgGridSequence.cpp
wwkong: Wouldn't this approach be inefficient in the case where one or more of the basis values (over dimensions) is zero, though? 

For example, if at least two basis values for a given point are zero then the contribution towards the Jacobian is also zero, but we are still using `num_dimensions * num_output` operations for this case.

mkstoyanov: Think of the particle swarm algorithm that you implemented recently, that's a good example application of how a sparse grid surrogate will be used. Of the thousands of evaluations of the swarm particles, how many of those hit on the zero values and how many do not. The zeros appear only in the very rare case of evaluating exactly at a grid point.

Adding all the if-statements increases the cost of the vast majority of cases only to potentially speedup a few cases. Even then, branching vs. multiplying by zeros is not a clear win for the branching, it may end up being more expensive anyway.

SparseGrids/tsgGridSequence.cpp
wwkong: Changed.

SparseGrids/tsgGridSequence.cpp
wwkong: Fair enough. I thought this might be relevant when the user is evaluating the derivative of the surrogate at abscissas (or some components of an abscissa) in the grid. For that specific case, many of the jacobian computations would be wasted.

As an aside, I will be making another PR to make the same change on the global grids.

PR #672, opened 2022-04-08
Optimization of differentiation in global grids

wwkong:
This PR removes every if-statement inside the major for-loops used in `GridGlobal::getDifferentiationWeights()`. 

Note:

- The tolerances used in the grid tests have been slightly adjusted.
- The computation of the derivative cache in `CacheLagrangeDerivative::getLagrangeDerivative` now requires `O(num_points)` additional storage space instead of `O(1)` space as in the previous version. This is because we are not splitting the logic based on whether or not `x` is at a node.

### DIFF
--- a/SparseGrids/tsgCacheLagrange.hpp
+++ b/SparseGrids/tsgCacheLagrange.hpp
@@ -72,10 +72,8 @@ public:
      *   and the actual nodes with the pre-computed Lagrange coefficients
      * - \b holds the coordinates of the canonical point to cache
      */
-    CacheLagrange(int num_dimensions, const std::vector<int> &max_levels, const OneDimensionalWrapper &rule, const double x[]){
-        cache.resize(num_dimensions);
-        offsets = rule.getPointsCount();
-
+    CacheLagrange(int num_dimensions, const std::vector<int> &max_levels, const OneDimensionalWrapper &rule, const double x[]) :
+            cache(std::vector<std::vector<T>>(num_dimensions, std::vector<T>())), offsets(rule.getPointsCount()){
         for(int dim=0; dim<num_dimensions; dim++){
             cache[dim].resize(offsets[max_levels[dim] + 1]);
             for(int level=0; level <= max_levels[dim]; level++)
@@ -138,9 +136,8 @@ public:
      *   and the actual nodes with the pre-computed Lagrange coefficients
      * - \b holds the coordinates of the canonical point to cache
      */
-    CacheLagrangeDerivative(int num_dimensions, const std::vector<int> &max_levels, const OneDimensionalWrapper &rule, const double x[]) {
-        offsets = rule.getPointsCount();
-        cache.resize(num_dimensions);
+    CacheLagrangeDerivative(int num_dimensions, const std::vector<int> &max_levels, const OneDimensionalWrapper &rule, const double x[]) :
+            cache(std::vector<std::vector<T>>(num_dimensions, std::vector<T>())), offsets(rule.getPointsCount()) {
         for(int dim=0; dim<num_dimensions; dim++){
             cache[dim].resize(offsets[max_levels[dim] + 1]);
             for(int level=0; level <= max_levels[dim]; level++)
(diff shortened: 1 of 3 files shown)


SparseGrids/tsgCacheLagrange.hpp
```diff
@@ -138,101 +138,58 @@ class CacheLagrangeDerivative : public CacheLagrange<T> {
...
-
-        // Check to see if x is one of the cached roots.
-        int match_idx = -1;
-        for (int j=0; j<num_points; j++) {
-            if (std::fabs(x - nodes[j]) <= Maths::num_tol) {
-                match_idx = j;
-                break;
-            }
+        // cc first stores fj'(x), aux_f stores fj(x), and aux_g stores gj(x).
+        T aux_f[num_points], aux_g[num_points];
```
mkstoyanov, line 168: You cannot do this. There is a GCC extension that allows the declaration of an array with length that depends on a runtime variable, but it is not allowed in the C++ standard. Use `std::vector<T>`

SparseGrids/tsgCacheLagrange.hpp
```diff
@@ -138,101 +138,58 @@ class CacheLagrangeDerivative : public CacheLagrange<T> {
      *   and the actual nodes with the pre-computed Lagrange coefficients
      * - \b holds the coordinates of the canonical point to cache
      */
-    CacheLagrangeDerivative(int num_dimensions, const std::vector<int> &max_levels, const OneDimensionalWrapper &rule, const double x[]) :
-            CacheLagrange<T>(num_dimensions, max_levels, rule, x) {
-        cacheDerivatives.resize(num_dimensions);
+    CacheLagrangeDerivative(int num_dimensions, const std::vector<int> &max_levels, const OneDimensionalWrapper &rule, const double x[]) {
+        offsets = rule.getPointsCount();
```
mkstoyanov, line 142: * initialize the `offsets` and `cache` inline to skip the default initialization, e.g., ` : offsets(rule.getPointsCount())`

SparseGrids/tsgCacheLagrange.hpp
wwkong: Changed.

SparseGrids/tsgCacheLagrange.hpp
wwkong: Changed.

SparseGrids/tsgCacheLagrange.hpp
```diff
@@ -139,100 +139,56 @@ class CacheLagrangeDerivative : public CacheLagrange<T> {
      * - \b holds the coordinates of the canonical point to cache
      */
     CacheLagrangeDerivative(int num_dimensions, const std::vector<int> &max_levels, const OneDimensionalWrapper &rule, const double x[]) :
-            CacheLagrange<T>(num_dimensions, max_levels, rule, x) {
-        cacheDerivatives.resize(num_dimensions);
+            cache(std::vector<std::vector<T>>(num_dimensions)), offsets(rule.getPointsCount()) {
         for(int dim=0; dim<num_dimensions; dim++){
-            cacheDerivatives[dim].resize(offsets[max_levels[dim] + 1]);
+            cache[dim].resize(offsets[max_levels[dim] + 1]);
```
mkstoyanov, line 144: * initialize that one too

wwkong: Do you mean I should initialize `cache` as follows?
```
cache(std::vector<std::vector<T>>(num_dimensions, std::vector<T>()))
```
Also, should I also apply the same changes to the `CacheLagrange` class?

mkstoyanov: Yes, I missed the `CacheLagrange` class back when I was coding it.

Inline initialization is better than initialize and then reinitialize, whenever reasonable and possible.

wwkong: Okay. I finished making the suggested changes.

PR #673, opened 2022-04-09
Add to differentiation and its tests to local polynomial grids

wwkong:


SparseGrids/tsgRuleLocalPolynomial.hpp
```diff
@@ -261,6 +265,86 @@ class templRuleLocalPolynomial : public BaseRuleLocalPolynomial{
...
+            // Note that the support will be of the form [a, b), except for x = +1.0, i.e., the rightmost node in the domain. This
+            // is to avoid double-counting derivatives at points of discontinuity.
+            double xn = scaleX(point, x);
+            bool rightmost_node = (x == 1.0 and xn == 1.0);
+            isSupported = (-1.0 <= xn and xn < 1.0) or rightmost_node;
+            if (isSupported) {
+                double an = scaleDiffX(point);
+                if (rule != rule_semilocalp and max_order == 1) {
+                    // Edge cases due to the logic of avoiding double-counting.
+                    if (x == 1.0) {
```
mkstoyanov, line 327: The first three if-statements are not needed, you would get the exact same result if you just apply the return formula on line `337`. I'm not sure about the 4th statement, but I wonder why that would be needed.

SparseGrids/tsgRuleLocalPolynomial.hpp
```diff
@@ -261,6 +265,86 @@ class templRuleLocalPolynomial : public BaseRuleLocalPolynomial{
...
+                }
+                if (rule == rule_semilocalp) {
+                    if (point == 1) return x - 0.5;
+                    if (point == 2) return x + 0.5;
+                }
+            }
+            // Note that the support will be of the form [a, b), except for x = +1.0, i.e., the rightmost node in the domain. This
+            // is to avoid double-counting derivatives at points of discontinuity.
+            double xn = scaleX(point, x);
+            bool rightmost_node = (x == 1.0 and xn == 1.0);
```
mkstoyanov, line 321: If `x == 1.0` then how can it be that `xn != 1.0`? The if-condition seems redundant.

SparseGrids/tsgRuleLocalPolynomial.hpp
```diff
@@ -364,6 +463,79 @@ class templRuleLocalPolynomial : public BaseRuleLocalPolynomial{
...
+            if (point == 0) return -1.0;
+            if (point == 1) return 1.0;
+            if (point == 2) return -2.0 * x;
+        } else if (rule == rule_localp0) {
+            if (point == 0) return -2.0 * x;
+        }
+        return (point % 2 == 0) ? 1.0 / 3.0 - x * (x + 2.0) : -1.0 / 3.0 + x * (x - 2.0);
+    }
+    double diffPWPower(int point, double x) const{
+        if (rule == rule_localp)     if (point <= 8) return diffPWCubic(point, x);
```
mkstoyanov, line 494: Why nest two if-statements when you can just use `and` logic.

SparseGrids/tsgRuleLocalPolynomial.hpp
```diff
@@ -364,6 +463,79 @@ class templRuleLocalPolynomial : public BaseRuleLocalPolynomial{
...
+        if (rule == rule_localpb)    if (point <= 4) return diffPWCubic(point, x);
+        if (rule == rule_localp0)    if (point <= 2) return diffPWCubic(point, x);
+        int level = getLevel(point);
+        int max_ancestors;
+        if (rule == rule_localp)     max_ancestors = level-2;
+        if (rule == rule_semilocalp) max_ancestors = level-1;
+        if (rule == rule_localpb)    max_ancestors = level-1;
+        if (rule == rule_localp0)    max_ancestors = level;
+        if (max_order > 0) max_ancestors = std::min(max_ancestors, max_order - 2);
+        int most_turns = 1;
```
mkstoyanov, line 505: Add a new line above this, since this is code with separate logic, makes it easier to see the logic of the lambda.

Comment on what the lambda does, e.g., captures by reference `most_turns` and `phantom_distance`, uses those as internal state variables, on each call it returns the next normalized ancestor node.

SparseGrids/tsgGridLocalPolynomial.cpp
```diff
@@ -846,6 +872,48 @@ double GridLocalPolynomial::evalBasisSupported(const int point[], const double x
     return f;
 }
 
+std::vector<double> GridLocalPolynomial::diffBasisRaw(const int point[], const double x[]) const{
```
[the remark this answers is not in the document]
mkstoyanov: Just noticed another problem, I don't think this is ever used.

SparseGrids/tsgGridLocalPolynomial.cpp
```diff
@@ -846,6 +872,48 @@ double GridLocalPolynomial::evalBasisSupported(const int point[], const double x
...
+        df[idx_zero_value] += nonzero_values_prod * rule->diffRaw(point[idx_zero_value], x[idx_zero_value]);
+    return df;
+}
+std::vector<double> GridLocalPolynomial::diffBasisSupported(const int point[], const double x[], bool &isSupported) const{
+    isSupported = false;
+    std::vector<double> diff_values(num_dimensions, 1.0);
+    bool isDimSupported = false;
+    for(int k=0; k<num_dimensions; k++) {
+        for(int j=0; j<k; j++) {
+            diff_values[k] *= rule->evalSupport(point[j], x[j], isDimSupported);
```
mkstoyanov, line 904: As you can see in the implementation of `evalSuppor()` the method is not trivial. The way you have ordered the for-loops, for different values of `k` you will have multiple calls to `evalSupport()` with identical values of `j`.

Make the outer loop over `j`, call `evalSuppor()` once, then have the two `k` loops to multiply the values that you need. You can compute the `rule->diffSupport()` either inside or in a separate loop.

SparseGrids/tsgGridLocalPolynomial.hpp
```diff
@@ -191,15 +198,32 @@ class GridLocalPolynomial : public BaseCanonicalGrid{
 
         for(const auto &r : roots){
             bool isSupported;
-            double basis_value = evalBasisSupported(work.getIndex(r), x, isSupported);
+            double basis_value;
```
mkstoyanov, line 201: Uninitialized `basis_value`. Initialize inline with `() ? : ` expression and set to zero for modes 3 and 4.

SparseGrids/tsgGridLocalPolynomial.hpp
```diff
@@ -181,6 +186,8 @@ class GridLocalPolynomial : public BaseCanonicalGrid{
      * - \b mode \b 0, ignore \b sindx and \b svals, find the non-zero basis functions multiply them by the surpluses and add to \b y
      * - \b mode \b 1, ignore \b y, form a sparse vector by std::vector::push_back() to the \b sindx and \b svals
      * - \b mode \b 2, same as \b mode \b 1 but it also sorts the entries within the vector (requirement of Nvidia cusparseDgemvi)
+     * - \b mode \b 3, same as \b mode \b 0, but replaces the non-zero basis function values with their derivatives
+     * - \b mode \b 4, same as \b mode \b 1, but replaces the non-zero basis function values with their derivatives
```
mkstoyanov, line 190: Do you need mode 4? We are not building a sparse matrix for the derivatives, this is used only in the case where we are computing the derivatives in batches and that's a whole other can of worms.

SparseGrids/tsgRuleLocalPolynomial.hpp
```diff
@@ -364,6 +463,79 @@ class templRuleLocalPolynomial : public BaseRuleLocalPolynomial{
...
+            most_turns *= 2;
+            phantom_distance = 2.0 * phantom_distance + 1.0;
+            int turns = (rule == rule_localp0) ? ((point+1) % most_turns) : ((point-1) % most_turns);
+            return (turns < most_turns / 2) ?
+                    (phantom_distance - 2.0 * ((double) turns)) :
+                    (-phantom_distance + 2.0 * ((double) (most_turns - 1 - turns)));
+        };
+
+        // Does not include the additional factor (1-x)(1+x) and the Lagrange coefficient.
+        std::vector<double> nodes(max_ancestors), left_prods(max_ancestors);
```
mkstoyanov, line 517: Is there a way to avoid allocating/deallocating vectors here?

SparseGrids/gridtestExternalTests.cpp
```diff
@@ -872,19 +876,26 @@ bool ExternalTester::testAllGlobal() const{
 bool ExternalTester::testLocalPolynomialRule(const BaseFunction *f, TasGrid::TypeOneDRule rule, const int depths[], const double tols[]) const{
     TasGrid::TasmanianSparseGrid grid;
     TestResults R;
-    TestType tests[3] = { type_integration, type_nodal_interpolation, type_internal_interpolation };
-    int orders[6] = { 0, 1, 2, 3, 4, -1 };
+    BaseFunction* const_fn = new OneOneP0();
```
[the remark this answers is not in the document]
wwkong: Changed. I added a new private member called `f11p0` to the `ExternalTester` class.

SparseGrids/gridtestExternalTests.cpp
```diff
@@ -1041,37 +1052,81 @@ bool ExternalTester::testAllPWLocal() const{
...
-    }else{
-        cout << setw(wfirst) << "Rule" << setw(wsecond) << IO::getRuleString(oned) << setw(wthird) << "FAIL" << endl; pass = false;
-    }}
+      const int depths1[30] = { 8, 8, 8, 8, 8,
+                                8, 8, 8, 8, 8,
+                                8, 8, 8, 8, 8,
+                                8, 8, 8, 8, 8,
+                                8, 8, 8, 8, 8,
+                                8, 8, 8, 8, 8};
+      const double tols1[30] = { 1.E-03, 5.E-01, 5.E-01, 5.E-00, 5.E-00,
```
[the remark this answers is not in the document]
wwkong: I changed the tolerance for the zeroth order test to `5E-01`, but this actually doesn't matter because I have a special case that only checks if the return derivative is zero (i.e, the tolerance does not impact this test).

I also added comments to each row of the tolerances, as requested.

SparseGrids/tsgGridLocalPolynomial.cpp
```diff
@@ -846,6 +872,48 @@ double GridLocalPolynomial::evalBasisSupported(const int point[], const double x
...
+        }
+    }
+    if (num_zero_values == 0)
+        for (int j=0; j<num_dimensions; j++)
+            df[j] += nonzero_values_prod / basis_values[j] * rule->diffRaw(point[j], x[j]);
+    else if (num_zero_values == 1)
+        df[idx_zero_value] += nonzero_values_prod * rule->diffRaw(point[idx_zero_value], x[idx_zero_value]);
+    return df;
+}
+std::vector<double> GridLocalPolynomial::diffBasisSupported(const int point[], const double x[], bool &isSupported) const{
```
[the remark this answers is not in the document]
wwkong: Changed. This function now writes to an input array.

SparseGrids/tsgRuleLocalPolynomial.hpp
wwkong: You're right. The first three are redundant, so I have removed them. 

We can see that the fourth one is necessary for `TasGrid::rule_localp`, `x == 1.0`, and `point == 2` by tracing the following facts: 

- `xn == 0.0` from the logic in `scaleX()`
- `an == 1.0` from the logic in `scaleDiffX()` and the value of `xn`
- `(xn >= 0 ? -1.0 : 1.0) * an == -1.0` from the above two statements
- at the boundary where `x == 1.0` only the left derivative exists 
- since the left derivative at `point == 2` for `xn == 0.0` is `1.0`, we should not take the logic in line `337` and hardcode it as `an`

SparseGrids/tsgRuleLocalPolynomial.hpp
wwkong: This happens in many cases. If `x == 1.0` and the support associated with `point` does not contain `1.0`, then `xn > 1.0`. So the bool `rightmost_node` is only true if `x` is the rightmost point in the domain associated with `point`.

SparseGrids/tsgRuleLocalPolynomial.hpp
wwkong: Changed. This was copied from `evalPWPower()`, and I'm always hesitant to modify the structure of existing code because I don't know if there is some hidden logic that I missed.

Should I also change other code segments where I see nested if-statements like this (e.g. `evalPWPower()` and `evalRaw()`)?

SparseGrids/tsgRuleLocalPolynomial.hpp
wwkong: Changed.

SparseGrids/tsgGridLocalPolynomial.cpp
wwkong: Changed. The calls to `rule->diffSupport()` are in a separate loop.

SparseGrids/tsgGridLocalPolynomial.hpp
wwkong: Changed.

SparseGrids/tsgRuleLocalPolynomial.hpp
wwkong: I avoided allocating `nodes` by using another lambda that inverts the operations made by `update_and_get_next_node()`, but I don't think you can avoid allocating `left_prods()` unless we allocate it outside of this function and pass it in as an input.

SparseGrids/tsgGridLocalPolynomial.hpp
wwkong: Mode 4 is needed in `GridLocalPolynomial::getDifferentiationWeights()`, which does solve a large, but sparse, triangular linear system of equations.

SparseGrids/gridtestExternalTests.cpp
```diff
@@ -56,7 +56,12 @@ std::vector<double> genRandom(int num_samples, int num_dimensions = 1){
 }
 double unitDerivativeTests(const BaseFunction *f, TasmanianSparseGrid &grid) {
     // Test the differentiate method on some common 1D nodes (in the 1st dimension) and return the max relative error.
-    std::vector<double> unique_nodes = {0.0, -1.0, 1.0, -0.5, 0.5, -0.5773502691896257, 0.5773502691896257};
+    std::vector<double> unique_nodes;
+    if (grid.getRule() == TasGrid::rule_localp0)
+        // Avoid boundary points as not all functions have value zero on these points.
+        unique_nodes = {0.0, -0.5, 0.5, -0.5773502691896257, 0.5773502691896257};
+    else
+        unique_nodes = {0.0, -1.0, 1.0, -0.5, 0.5, -0.5773502691896257, 0.5773502691896257};
```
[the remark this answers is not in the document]
mkstoyanov: That's correct. The zero-boundary rule should be used to test a zero-boundary function.

SparseGrids/tsgGridLocalPolynomial.hpp
mkstoyanov: I don't get it. The matrix that should be used here is the one from mode 1, there is no special separate matrix for the derivatives.

SparseGrids/tsgRuleLocalPolynomial.hpp
mkstoyanov: So your check is whether you're using the right-most basis function combined with the right-most domain point, one relates to `x` and the other to `point`. The name is a bit confusing, what is `rightmost` and how that relates to the support.

Why not combine with the next line:
```
isSupported = (-1.0 <= xn and xn < 1.0) or (xn == 1.0 and x == 1.0);
```
Now we can see that we consider the interval `xn \in [-1.0, 1.0)` but make an exception for the edge case when `x==1.0` we use `xn \in [-1.0, 1.0]`

wwkong: Sure. I've changed it to 
```
isSupported = (-1.0 <= xn and xn < 1.0) or (xn == 1.0 and x == 1.0);
```
in the latest push.

SparseGrids/tsgGridLocalPolynomial.hpp
wwkong: I'm not sure about that. Since we have `num_dimension` more variables to solve from, shouldn't the matrix have `num_dimension` times more entries than the one for mode 1? Also, shouldn't it be filled with the derivative values rather than the function values?

SparseGrids/gridtestExternalTests.cpp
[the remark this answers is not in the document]
wwkong: Okay. Let's keep this specific exception, then.

SparseGrids/tsgGridLocalPolynomial.hpp
mkstoyanov: You would only ever form the matrix here in the case of batch differentiation. We don't differentiate in batch.

The only matrix that we are working with is the one corresponding to the computation of surplus coefficients and that matrix does not depend on `x` and thus is a constant with respect to differentiation.

mkstoyanov: Remove case 4 and I think this is ready.

wwkong: Maybe I am missing something here conceptually. 

Suppose a real-valued interpolant `F` is given by `F(x) = <Φ(x), s>`, where `Φ(x)` is vector of basis functions values evaluated at `x` and `s` is the vector of surplus coefficients. As far as I can tell, `walkTree<1>()` is generating `Φ(x)` and directly depends on `x`. My implementation of `walkTree<4>()` is just generating `∇Φ(x)`, which clearly cannot be the same matrix/vector as `Φ(x)`.

Note that the application of the inverse transpose of the matrix associated with the surplus coefficients is done in the calls of `applyTransformationTransposed<T>()` and has nothing to do with `walkTree<T>()`.

mkstoyanov (approved):

PR #674, opened 2022-04-18
Add differentiation to Fourier grids + small bug fixes

wwkong:
Bug fixes include:

- Corrections to derivatives defined in `gridtestTestFunctions.cpp`
- Including the effect of affine (non-conformal) domain transforms in `TasmanianSparseGrid::differentiate()`

### DIFF
--- a/SparseGrids/TasmanianSparseGrid.cpp
+++ b/SparseGrids/TasmanianSparseGrid.cpp
@@ -391,7 +391,9 @@ void TasmanianSparseGrid::getDifferentiationWeights(const std::vector<double> &x
     getDifferentiationWeights(x.data(), weights.data());
 }
 void TasmanianSparseGrid::getDifferentiationWeights(const double x[], double weights[]) const{
+    // See differentiate() for how the domain transforms are taken into consideration.
     Data2D<double> x_tmp;
+    // Jacobian of f(.) at g(x).
     if (isGlobal()) {
         get<GridGlobal>()->getDifferentiationWeights(formCanonicalPoints(x, x_tmp, 1), weights);
     } else if (isSequence()) {
@@ -402,6 +404,16 @@ void TasmanianSparseGrid::getDifferentiationWeights(const double x[], double wei
         get<GridFourier>()->getDifferentiationWeights(formCanonicalPoints(x, x_tmp, 1), weights);
     } else {
         throw std::runtime_error("ERROR: getDifferentiationWeights() cannot be called for grids of this type");
+    }
+    // Jacobian of f(g(.)) at x.
+    if (domain_transform_a.size() != 0 or conformal_asin_power.size() != 0) {
+        int num_dimensions = getNumDimensions();
+        int num_points = getNumPoints();
+        std::vector<double> jacobian_g_diag(num_dimensions);
+        diffCanonicalTransform<double>(x, jacobian_g_diag.data());
+        for(int i=0; i<num_points; i++)
+            for(int j=0; j<num_dimensions; j++)
+                weights[i * num_dimensions + j] = weights[i * num_dimensions + j] * jacobian_g_diag[j];
     }
 }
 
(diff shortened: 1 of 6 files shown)


SparseGrids/TasmanianSparseGrid.cpp
```diff
@@ -391,16 +391,30 @@ void TasmanianSparseGrid::getDifferentiationWeights(const std::vector<double> &x
...
         get<GridSequence>()->getDifferentiationWeights(formCanonicalPoints(x, x_tmp, 1), weights);
     } else if (isLocalPolynomial()) {
         get<GridLocalPolynomial>()->getDifferentiationWeights(formCanonicalPoints(x, x_tmp, 1), weights);
+    } else if (isFourier()) {
+        get<GridFourier>()->getDifferentiationWeights(formCanonicalPoints(x, x_tmp, 1), weights);
     } else {
         throw std::runtime_error("ERROR: getDifferentiationWeights() cannot be called for grids of this type");
     }
+    // Jacobian of f(g(.)) at x.
+    if (domain_transform_a.size() != 0 or conformal_asin_power.size() != 0) {
```
mkstoyanov, line 409: If you have unfinished features like the domain transforms, throw an exception and remove it late. Right now, someone may use this an get garbage results not realizing what's happening.

SparseGrids/TasmanianSparseGrid.cpp
```diff
@@ -479,15 +493,31 @@ void TasmanianSparseGrid::integrate(double q[]) const{
...
     } else if (isLocalPolynomial()) {
-        get<GridLocalPolynomial>()->differentiate(x, jacobian);
+        get<GridLocalPolynomial>()->differentiate(formCanonicalPoints(x, x_tmp, 1), jacobian);
+    } else if (isFourier()) {
+        get<GridFourier>()->differentiate(formCanonicalPoints(x, x_tmp, 1), jacobian);
     } else {
         throw std::runtime_error("ERROR: in differentiate(), jacobians/gradients are not available for this type of grid");
     }
+    // Jacobian of f(g(.)) at x.
+    if (domain_transform_a.size() != 0 or conformal_asin_power.size() != 0) {
```
mkstoyanov, line 512: Same here, throw and exception if a transform has been set.

SparseGrids/TasmanianSparseGrid.cpp
wwkong: This is handled by the new `TasmanianSparseGrid::diffCanonicalTransform()` function in this PR. Specifically, that function throws an exception if conformal transforms are in place, but not affine ones.

SparseGrids/TasmanianSparseGrid.cpp
wwkong: Same as above

SparseGrids/TasmanianSparseGrid.cpp
mkstoyanov: Why is the if-statement needed here?

SparseGrids/TasmanianSparseGrid.cpp
```diff
@@ -818,6 +848,31 @@ template<typename FloatType> const FloatType* TasmanianSparseGrid::formCanonical
 template const float* TasmanianSparseGrid::formCanonicalPoints(const float *x, Data2D<float> &x_temp, int num_x) const;
 template const double* TasmanianSparseGrid::formCanonicalPoints<double>(const double *x, Data2D<double> &x_temp, int num_x) const;
 
+template<typename FloatType> void TasmanianSparseGrid::diffCanonicalTransform(const FloatType*, FloatType *jacobian_diag) const {
```
mkstoyanov, line 851: * If the transform does not depend on `x`, then remove it. The first input is not needed.
* The way you're using this, you define a vector, initialize to zeros, then pass in the pointer, then reinitialize to 1.
    * declare the vector inside the method and return it

SparseGrids/TasmanianSparseGrid.cpp
```diff
@@ -818,6 +848,31 @@ template<typename FloatType> const FloatType* TasmanianSparseGrid::formCanonical
...
+        jacobian_diag[i] = 1.0;
+    if (conformal_asin_power.size() != 0) {
+        throw std::runtime_error("ERROR: in diffCanonicalTransform() derivatives/Jacobians are not available for conformal mappings");
+    }
+    if (domain_transform_a.size() != 0) {
+        // Based on the logic in mapTransformedToCanonical(). These are all affine transforms.
+        TypeOneDRule rule = base->getRule();
+        if (rule == rule_gausslaguerre or rule == rule_gausslaguerreodd)
+            for(int j=0; j<num_dimensions; j++)
+                jacobian_diag[j] *= (FloatType) domain_transform_b[j];
```
mkstoyanov, line 863: You initialize to `1.0` then you multiply one by the transform and then reassign.

If you use just the assignment `=`, then you will skip the multiplication and one extra `1.0`.

SparseGrids/gridtestTestFunctions.cpp
```diff
@@ -271,23 +270,27 @@ void TwoOne1DCurved::getDerivative(const double x[], double y[]) const{ y[0] = -
...
 }
 
 TwoOneConformalOne::TwoOneConformalOne(){} TwoOneConformalOne::~TwoOneConformalOne(){} int TwoOneConformalOne::getNumInputs() const{ return 2; } int TwoOneConformalOne::getNumOutputs() const{ return 1; }
 const char* TwoOneConformalOne::getDescription() const{ return "f(x) = 1 / ((1 + 5x^2)*(1 + 5y^2))"; }
 void TwoOneConformalOne::eval(const double x[], double y[]) const{ y[0] = 1.0 / ((1.0 + 5.0*x[0]*x[0]) * (1.0 + 5.0*x[1]*x[1])); } void TwoOneConformalOne::getIntegral(double y[]) const{ y[0] = 1.028825601981092*1.028825601981092; }
 void TwoOneConformalOne::getDerivative(const double x[], double y[]) const{
-    double a0 = (1.0 + 5.0*x[0]*x[0]) * (1.0 + 5.0*x[1]*x[1]);
-    y[0] = 10.0 * x[0] / (a0 * a0);
-    y[1] = 10.0 * x[1] / (a0 * a0);
+    double a0 = (1.0 + 5.0 * x[0] * x[0]) * (1.0 + 5.0 * x[1] * x[1]);
```
mkstoyanov, line 291: That's just style adjustments.

SparseGrids/TasmanianSparseGrid.cpp
wwkong: The if-statement is needed to avoid calling `TasmanianSparseGrid::diffCanonicalTransform()` when the available transformation data, e.g., `domain_transform_a`, is empty, since that function requires it to be non-empty. We also save on some allocations and excessive operations.

mkstoyanov: Read the code inside `diffCanonicalTransform()`, you are first checking if `conformal_asin_power` is not empty (and raising an exception), then checking if `domain_transform_a` is not empty and doing things. This if-statement looks totally redundant.

mkstoyanov: On a related note, `conformal_asin_power.size() != 0` works, but sometimes it is more expressive to use
```
if (not domain_transform_a.empty() or not conformal_asin_power.empty())
```

SparseGrids/TasmanianSparseGrid.cpp
wwkong: Good points. I've fixed this in the latest PR.

SparseGrids/gridtestTestFunctions.cpp
wwkong: Changed back.

SparseGrids/TasmanianSparseGrid.cpp
wwkong: Ah, I missed that. In that case, I have changed the if-statement to 
```
if (not domain_transform_a.empty())
```
since we want to avoid the allocation in `TasmanianSparseGrid::diffCanonicalTransform()` and the expensive double for-loop in the if-block when `domain_transform_a` is empty, i.e., when we don't need to scale the canonical Jacobian.

PR #676, opened 2022-04-28
Add differentiation to Wavelet grids

wwkong:


SparseGrids/gridtestExternalTests.cpp
```diff
@@ -288,7 +288,7 @@ TestResults ExternalTester::getError(const BaseFunction *f, TasGrid::TasmanianSp
         };
         R.error = err;
     }else if (type == type_internal_interpolation or type == type_internal_differentiation){
-        if (type == type_internal_differentiation and !(grid.isGlobal() or grid.isSequence() or grid.isLocalPolynomial() or grid.isFourier())) {
+      if (type == type_internal_differentiation and !(grid.isGlobal() or grid.isSequence() or grid.isLocalPolynomial() or grid.isWavelet() or grid.isFourier())) {
```
mkstoyanov, line 291: Alignment is off.

SparseGrids/tsgGridWavelet.cpp
```diff
@@ -143,6 +143,25 @@ void GridWavelet::getInterpolationWeights(const double x[], double *weights) con
     if (inter_matrix.getNumRows() != num_points) buildInterpolationMatrix();
     inter_matrix.invertTransposed(acceleration, weights);
 }
+void GridWavelet::getDifferentiationWeights(const double x[], double weights[]) const {
+    const MultiIndexSet &work = (points.empty()) ? needed : points;
+    int num_points = work.getNumIndexes();
+    #pragma omp parallel for
+    for (int i=0; i<num_points; i++) {
+      evalDiffBasis(work.getIndex(i), x, &(weights[i * num_dimensions]));
```
mkstoyanov, line 151: alignment

mkstoyanov: check the whole funciton

SparseGrids/tsgGridWavelet.cpp
```diff
@@ -331,6 +366,32 @@ double GridWavelet::evalIntegral(const int p[]) const{
     }
     return v;
 }
+void GridWavelet::evalDiffBasis(const int p[], const double x[], double jacobian[]) const {
+    // Evaluates the derivative of the wavelet basis given at point p at the coordinates given by x.
+    // The derivative function of a symmetric function will be antisymmetric across the origin for every x != 0.0.
+    std::vector<double> value_cache(num_dimensions), derivative_cache(num_dimensions);
+    for(int i=0; i<num_dimensions; i++) {
+        value_cache[i] = rule1D.eval<0>(p[i], x[i]);
+        derivative_cache[i] = rule1D.eval<1>(p[i], x[i]);
```
mkstoyanov, line 375: Derivative cache is used only once, why cache it instead of writing to `jacobian` directly. The second loop down will only deal with the contribution of the `value_cache`.

SparseGrids/gridtestExternalTests.cpp
wwkong: Fixed.

SparseGrids/tsgGridWavelet.cpp
wwkong: Fixed.

SparseGrids/tsgGridWavelet.cpp
```diff
@@ -331,6 +366,32 @@ double GridWavelet::evalIntegral(const int p[]) const{
...
+    }
+
+    std::fill(jacobian, jacobian + num_dimensions, 0.0);
```
mkstoyanov, line 378: use `std::fill_n()`

SparseGrids/tsgGridWavelet.cpp
```diff
@@ -331,6 +366,32 @@ double GridWavelet::evalIntegral(const int p[]) const{
...
+    // Note that excessive branching is done to mirror the logic in evalBasis() and evalIntegral().
+    for(int i=0; i<num_dimensions; i++) {
```
mkstoyanov, line 380: You have nested loops that perform `num_dimensions^2` operations and even with the if-statements and the breaks, the complexity is bounded below by `num_dimensions`.

You can guarantee complexity of `num_dimensions` with the following:
```
double t = 1.0;
for(int i=1; i<num_dimensions; i++){
    t *= value_cache[i];
    jacobian[i] *= t;
}
t = 1.0;
for(int i=num_dimensions-1; i>=0; i--){
    t *= value_cache[i];
    jacobian[i] *= t;
}
```
We can break the first loop when `t` becomes zero and zero out the remainder of the jacobian and we can shorten the second loop to only the non-negative portions (and zero out if `t` hits zero), but the zero-out will still require operations and the current set assumes that we don't have `derivative_cache`, instead the derivative values are already written into the Jacobian.

SparseGrids/tsgRuleWavelet.cpp
```diff
@@ -295,35 +295,53 @@ double RuleWavelet::getWeight(int point) const{
...
+            sgn = -1.0;
             x = -x;
         }else if(point == 4){
             point = 3;
+            sgn = -1.0;
             x = -x;
         }
         const double *phi = &(data[1][((point+1)/2) * num_data_points]);
-        return interpolate(phi, x);
+        return (mode == 0 ? 1.0 : sgn) * interpolate<mode>(phi, x);
```
mkstoyanov, line 344: In `mode == 0` this still multiplies by 1. It will probably get optimized out, but still, an explicit way is:
```
return (mode == 0 ? interpolate<mode>(phi, x) : sgn * interpolate<mode>(phi, x));
```
Note that this will never evaluate `interpolate` twice regardless of the value of `mode`.

SparseGrids/tsgRuleWavelet.cpp
```diff
@@ -332,34 +350,41 @@ inline double RuleWavelet::eval_cubic(int point, double x) const{
...
             point = 25 - point;
         }
         point -= 9;
-        return interpolate(&data[3][point*num_data_points],x);
+        return (mode == 0 ? 1.0 : sgn) * interpolate<mode>(&data[3][point*num_data_points],x);
     }
     // Standard lifted wavelets.
     int subindex = (point - 1) % (1 << l);
     double scale = pow(2,l-4);
-    // Left Boundary
```
mkstoyanov, line 352: Is this no longer related to the `Left Boundary`?

SparseGrids/tsgGridWavelet.cpp
wwkong: Removed the cache in the latest push.

SparseGrids/tsgGridWavelet.cpp
wwkong: Removed this line due to the next comment.

SparseGrids/tsgGridWavelet.cpp
wwkong: Sounds good. I changed this in the latest push to include the zero-ing out logic. 

I do think that it should be `value_cache[i-1]` in the first block, `value_cache[i+1]` in the second block, and `for(int i=num_dimensions-2; i>=0; i--)` for the second for-loop, though.

SparseGrids/tsgRuleWavelet.cpp
wwkong: Changed.

SparseGrids/tsgRuleWavelet.cpp
wwkong: Each boundary type is now in its own if/else if/else block, so I moved this comment downwards.

mkstoyanov: OK!

SparseGrids/tsgRuleWavelet.cpp
```diff
@@ -390,51 +415,77 @@ double RuleWavelet::getSupport(int point) const{
...
-    if ((x <= -0.5)){
-        return -0.5 * x - 0.5;
-    }else if (x >= 0.0){
-        return 0.5 * x - 0.25;
-    }else if (x <= -0.25){
-        return 4.0 * x + 1.75;
-    }else{
-        return -4.0 * x - 0.25;
+    // The ordering is important.
+    if ((right and x < -0.5) or (!right and x <= -0.5)) {
```
mkstoyanov, line 474: This is a bit messy, the `right` and `!right` are almost identical. How about:
```
if (x < -0.5 or (x == -0.5 and not right))
```
I also don't like that the messy right-not-right logic will come into play in both modes. I would like to have one section for `mode == 0` that does the old logic without the messy `!right` and a separate differentiation section that does the special case work.

SparseGrids/tsgGridWavelet.cpp
```diff
@@ -331,6 +366,37 @@ double GridWavelet::evalIntegral(const int p[]) const{
...
+        jacobian[i] = rule1D.eval<1>(p[i], x[i]);
+    }
+
+    // Note that excessive branching is done to mirror the logic in evalBasis() and evalIntegral().
+    int back_idx = num_dimensions-2;
+    double t = 1.0;
+    for(int i=1; i<num_dimensions; i++){
+        t *= value_cache[i-1];
+        jacobian[i] *= t;
+        if (t == 0.0) {
```
mkstoyanov, line 384: If `t` becomes zero in the first loop, you need to zero out `jacobian` for the entries after the `back_index`.

In the case of eval, breaking on a zero value will save several calls to `evalBasis()` and each call consists of multiple if-statements and index and floating point operations. If you break early here, you will save on couple of multiplications, which is cheap already. We can (and should) add differentiation to the sparse grid benchmark to test this, but I think the early break will not help and may even hurt performance.

I haven't done any PRs lately, let me add differentiation to the benchmark so we can test this. We can talk about this later in a call.

SparseGrids/tsgRuleWavelet.cpp
```diff
@@ -332,34 +350,41 @@ inline double RuleWavelet::eval_cubic(int point, double x) const{
             // i.e. 7 or 8
             // These wavelets are reflections across the y-axis of 6 & 5, respectively
             x = -x;
+            sgn = -1.0;
             point = 13 - point;
         }
         point -= 5;
-        return interpolate(&data[2][point*num_data_points],x);
+        return (mode == 0 ? 1.0 : sgn) * interpolate<mode>(&data[2][point*num_data_points],x);
```
mkstoyanov, line 357: You have a whole bunch of those, you changed only one above. This one still has the extra multiplication.

SparseGrids/tsgGridWavelet.cpp
mkstoyanov: Read the long paragraph, in the other two cases you're saving big, here you're saving very little if anything at all.

Wait for the benchmark and we will test it.

mkstoyanov: I see a lot of run-to-run variability when using 200 evals. Removing the early breaks does seem to make things a bit better, though.

I'll bump to 1000 repetitions.

mkstoyanov: Same result, I see a slight bump in performance if I remove the early breaking.

wwkong: Gotcha. I've removed the early breaking in the latest push.

wwkong: Okay. Let me run it a few times.

wwkong: I've tested it a few times on my home PC with an Intel(R) Core(TM) i7-10700K CPU, and there is only the slightest margin of improvement if I remove early breaking.

Here is a sample output with breaking:
```
./benchmarksgrid differentiate wavelet 4 1 2 level rleja 3 2000 5
differentiate wavelet 4 1 2 level rleja 3 2000 5 (points: 9025)
    note: reporting total time (ms), does not normalize (divide) by iterations
   3374
   3416
   3448
   3758
   4262
```
and here is one without breaking:
```
./benchmarksgrid differentiate wavelet 4 1 2 level rleja 3 2000 5
differentiate wavelet 4 1 2 level rleja 3 2000 5 (points: 9025)
    note: reporting total time (ms), does not normalize (divide) by iterations
   3246
   3327
   3473
   3745
   4225
```
Based on the results, I'll leave out the early breaking.

SparseGrids/tsgGridWavelet.cpp
```diff
@@ -331,6 +366,25 @@ double GridWavelet::evalIntegral(const int p[]) const{
     }
     return v;
 }
+void GridWavelet::evalDiffBasis(const int p[], const double x[], double jacobian[]) const {
+    // Evaluates the derivative of the wavelet basis given at point p at the coordinates given by x.
+    // The derivative function of a symmetric function will be antisymmetric across the origin for every x != 0.0.
+    std::vector<double> value_cache(num_dimensions), derivative_cache(num_dimensions);
```
mkstoyanov, line 372: * remove `derivative_cache`

SparseGrids/tsgRuleWavelet.cpp
```diff
@@ -390,52 +417,89 @@ double RuleWavelet::getSupport(int point) const{
...
+    // If `right` is true, then we return the right derivatives/values (if they exist); else, we return the left ones.
     if (std::abs(x + 0.5) > 0.5) return 0.0;
 
-    if ((x <= -0.75)){
-        return 0.75 * (7.0 * x + 6.0);
-    }else if (x <= -0.5){
-        return -0.25 * (11.0 * x + 6.0);
-    }else{
-        return 0.25 * x;
+    if ((right and x < -0.75) or (!right and x <= -0.75)) {
```
mkstoyanov, line 454: * right and not-right

SparseGrids/tsgGridWavelet.cpp
```diff
@@ -313,11 +332,27 @@ void GridWavelet::integrate(double q[], double *conformal_correction) const{
     }
 }
 
+void GridWavelet::differentiate(const double x[], double jacobian[]) const{
+    std::fill(jacobian, jacobian + num_outputs * num_dimensions, 0.0);
```
mkstoyanov, line 336: * use `fill()` when you have two iterators, e.g., `fill(x.begin(), x.end(), 0.0)` or `fill(x, y, 1.0)`
* use `fill_n()` when you have a range, e.g., `fill_n(x, x + num, 0.0)` or `fill_n(x.begin(), x.begin() + num, 1.0)`
* do not user, `fill(x, x + num, 1.0)`

mkstoyanov: * do not use `fill_n(x.begin(), x.size(), 1.0)`, prefer `fill(x.begin(), x.end(), 1.0)`

mkstoyanov (changes requested):

SparseGrids/tsgGridWavelet.cpp
wwkong: Done.

SparseGrids/tsgRuleWavelet.cpp
wwkong: Changed.

SparseGrids/tsgGridWavelet.cpp
wwkong: Changed.

SparseGrids/tsgGridWavelet.cpp
mkstoyanov: Tested on Ryzen, no change detected. If there's no difference, then we keep the simpler code.

The CI system has an issue, I will manually test tomorrow and merge.

PR #677, opened 2022-05-05
Add Python wrappers for the Particle Swarm state/algorithm

wwkong:


wwkong:
Removing the explicit cast to `c_void_p` to the Python state pointer `ParticleSwarmState.pStatePtr` when it is passed in as an argument always results in a segfault. I'll try to see why this is the case.

mkstoyanov:
* when you deal with pointers, try using hexadecimal, it makes a lot more sense than the binary format
* after making a bunch of experiments similar to you, it appears that there is some truncation of the value
* I have not idea why that would manifest here and not in TasmanianSG
    * people (including me) have used TasmanianSG for over 5 years on all sorts of machine
* we have to move one and work around this issue, hopefully we'll figure it out down the line
* moving to other comments

InterfacePython/TasmanianParticleSwarm.py
```diff
@@ -0,0 +1,264 @@
...
+    '''
+    def __init__(self, iNumDimensions, iNumParticles):
+        '''
+        Constructs a new state with a given number of dimensions and particles.
+
+        iNumDimensions : positive integer indicating the number of dimensions.
+        iNumParticles  : positive integer indicating the number of particles.
+        '''
+        self.TasmanianParticleSwarmState = True
+        self.pStatePntr = pLibDTSG.tsgParticleSwarmState_Construct(iNumDimensions, iNumParticles)
```
mkstoyanov, line 85: Whatever the `c_void_p` issue is here, the proper solution is to change this to:
```
self.pStatePntr = c_void_p( pLibDTSG.tsgParticleSwarmState_Construct(iNumDimensions, iNumParticles) )
```
then you don't have to type `c_void_p` in every call down the line.

DREAM/Optimization/TasmanianOptimizationWrapC.cpp
```diff
@@ -0,0 +1,150 @@
...
+    void tsgParticleSwarmState_ClearCache(void* state) {
+        reinterpret_cast<ParticleSwarmState*>(state)->clearCache();
+    }
+    void tsgParticleSwarmState_InitializeParticlesInsideBox(void* state, const double box_lower[], const double box_upper[],
+                                                            const char* random_type, const int random_seed, tsg_dream_random random_callback) {
+
+        // Create the U[0,1] random number generator.
+        std::minstd_rand park_miller((random_seed == -1) ? static_cast<long unsigned>(std::time(nullptr)) : random_seed);
+        std::uniform_real_distribution<double> unif(0.0, 1.0);
+        srand((unsigned int) ((random_seed == -1) ? static_cast<long unsigned>(std::time(nullptr)) : random_seed));
```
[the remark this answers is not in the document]
wwkong: Changed.

DREAM/Optimization/tsgParticleSwarm.hpp
```diff
@@ -86,53 +86,63 @@ class ParticleSwarmState {
...
     //! \brief Return the previously best known position in the swarm.
+    inline void getBestPosition(double bp[]) const {std::copy_n(best_particle_positions.begin() + num_particles * num_dimensions, num_dimensions, bp);}
     inline std::vector<double> getBestPosition() const {
         std::vector<double> best_position(num_dimensions);
-        std::copy_n(best_particle_positions.begin() + num_particles * num_dimensions, num_dimensions, best_position.begin());
+        getBestPosition(best_position.data());
         return best_position;
     }
     //! \brief Return the state vector.
+    inline void getStateVector(bool sv[]) const {
```
mkstoyanov, line 105: Is this ever used? The returned `std::vector<bool>` is used only for testing, testing is fine, but why is this needed?

If you want to expose the state to the user, you should use expressive call, e.g., `isPositionInitialized()`.

InterfacePython/Tasmanian.py
```diff
@@ -41,6 +41,7 @@
 SparseGrid = TasmanianSparseGrid
 
 import TasmanianDREAM as DREAM
+import TasmanianOPT as OPT
```
mkstoyanov, line 44: Naming conventions:

TasmanianSG.py and TasmanianOPT.py are fine

But we  should have:
```
import TasmanianOPT as Optimization
```
which is more expressive.

InterfacePython/TasmanianParticleSwarm.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def getNumParticles(self):
+        '''
+        Return the number of particles.
+        '''
+        return pLibDTSG.tsgParticleSwarmState_GetNumParticles(self.pStatePntr)
+
+    def getParticlePositions(self):
+        '''
+        Return the particle positions as a NumPy array.
```
mkstoyanov, line 107: Describe the array, i.e., it is a 2D array with rows corresponding to one thing and columns corresponding to another.

InterfacePython/TasmanianParticleSwarm.py
```diff
@@ -0,0 +1,264 @@
...
+    oParticleSwarmState : an instance of the Python ParticleSwarmState class; it will contain the results of applying the algorithm.
+    fInertiaWeight      : a double that controls the speed of the particles; should be in (0,1).
+    fCognitiveCoeff     : a double that controls how much a particle favors its own trajectory; usually in [1,3].
+    fSocialCoeff        : a double that controls how much a particle favors the swarm's trajectories; usually in [1,3].
+    random01            : a DREAM.RandomGenerator instance that produces floats in [0,1]
+
+    NOTE: np.frombuffer() is used below in order to avoid copying the data owned by the C pointers into the generated NumPy arrays.
+    '''
+    def cpp_obj_fn(num_dim, num_batch, x_batch_ptr, fval_ptr):
+        py_x_batch_ptr = cast(x_batch_ptr, POINTER(c_double * num_batch * num_dim))
```
mkstoyanov, line 249: The calls to `cast()` `frombuffer()` and `resize()` can be combined into a single call to:
```
aX = np.ctypeslib.as_array(x_batch_ptr, (num_batch, num_dim))
```

DREAM/Optimization/TasmanianOptimizationWrapC.cpp
```diff
@@ -0,0 +1,150 @@
...
+#ifndef __TASMANIAN_OPTIMIZATION_WRAPC_CPP
+#define __TASMANIAN_OPTIMIZATION_WRAPC_CPP
+
+#include "tsgParticleSwarm.hpp"
+
+// --------------------------- C Interface for use with Python ctypes and potentially other C codes --------------------------- //
+
+// C Function Pointer Aliases
+using tsg_dream_random = double (*)();
+using tsg_optim_dom_fn = bool   (*)(const int, const double[]);
```
mkstoyanov, line 43: These should be signatures to C functions, the C programming language doesn't have a standard `bool` variable. Use int instead with the convention that `0` is false and `not 0` is true.

InterfacePython/TasmanianParticleSwarm.py
```diff
@@ -0,0 +1,264 @@
...
+
+from TasmanianConfig import __path_libdream__
+from TasmanianConfig import TasmanianInputError as InputError
+
+import TasmanianDREAM as DREAM
+
+c_double_p        = POINTER(c_double)
+np_double_arr     = ndpointer(c_double, flags="C_CONTIGUOUS")
+type_optim_obj_fn = CFUNCTYPE(None, c_int, c_int, c_double_p, c_double_p)
+type_optim_dom_fn = CFUNCTYPE(c_bool, c_int, c_double_p)
```
mkstoyanov, line 45: I cannot see anything that guarantees that `c_bool` in python, `_Bool` in C and `bool` in C++ are the same.

I would use integers and convert to the boolean of Python or C++ as needed.

InterfacePython/TasmanianParticleSwarm.py
```diff
@@ -0,0 +1,264 @@
...
+        pLibDTSG.tsgParticleSwarmState_ClearCache(self.pStatePntr)
+
+    def initializeParticlesInsideBox(self, lfBoxLower, lfBoxUpper, random01 = DREAM.RandomGenerator(sType = "default")):
+        '''
+        Initialize the particle swarm state with randomized particles (determined by gen_random01) inside a box bounded by the
+        parameters box_lower and box_upper. The i-th entry in these parameters respectively represent the lower and upper
+        bounds on the particle position values for the i-th dimension. The parameter random01 controls the distribution of
+        the particles.
+
+        lfBoxLower : a one-dimensional NumPy array with lfBoxLower.shape[0] = .getNumDimensions()
```
mkstoyanov, line 212: Why the `.` notation, the method is either `self.getNumDimensions()` or simply `getNumDimensions()`.

Makefile
```diff
@@ -239,7 +239,6 @@ clean:
 	rm -fr include
 	rm -fr ./SparseGrids/*.o
 	rm -fr ./DREAM/*.o
-	rm -fr ./DREAM/Optimization/*.o
```
mkstoyanov, line 242: Why deleting this? It was needed for cleanup after the compilation process, the object files are in `./DREAM/Optimization/*.o`

InterfacePython/TasmanianParticleSwarm.py
wwkong: This is to stay consistent with the other DREAM modules (e.g., see the documentation for `TasmanianDreamState.setState()`). I'll fix it to one of the above in the next push.

DREAM/Optimization/tsgParticleSwarm.hpp
wwkong: I decided to make additional functions (`isPositionInitialized()`, `isVelocityInitialized()`, `isBestPositionInitialized()`, `isCacheInitialized()`)  to expose these states to the user.

InterfacePython/Tasmanian.py
wwkong: Changed.

InterfacePython/TasmanianParticleSwarm.py
wwkong: Changed here and in similar functions.

InterfacePython/TasmanianParticleSwarm.py
wwkong: Changed here and in the other lambda in this function.

DREAM/Optimization/TasmanianOptimizationWrapC.cpp
wwkong: Changed.

InterfacePython/TasmanianParticleSwarm.py
wwkong: Changed. I've used integers for the function specifications and applied casting to a Boolean when necessary.

InterfacePython/TasmanianParticleSwarm.py
wwkong: Done.

Makefile
wwkong: This was a mistake. I was playing around with Makefile and forgot to put this line back. It should be correct in the latest push.

InterfacePython/TasmanianParticleSwarm.py
```diff
@@ -0,0 +1,295 @@
...
+import numpy as np
+import sys
+
+from TasmanianConfig import __path_libdream__
+from TasmanianConfig import TasmanianInputError as InputError
+
+import TasmanianDREAM as DREAM
+
+c_double_p        = POINTER(c_double)
+np_double_arr     = ndpointer(c_double, flags="C_CONTIGUOUS")
```
mkstoyanov, line 43: Is this needed? We are using arrays that have a logical 2D format, but we don't know the dimensions prior to any calls, so we cannot realistically enforce those.

We should stick to `POINTER(c_double)`

InterfacePython/TasmanianParticleSwarm.py
```diff
@@ -0,0 +1,295 @@
...
+type_optim_obj_fn = CFUNCTYPE(None, c_int, c_int, c_double_p, c_double_p)
```
mkstoyanov, line 44: You can state `POINTER(c_double)` instead of making a variable `c_double_p`.

The variable doesn't really save on coding, since type definitions don't have logic or arithmetic, on the other hand, adding the new type means that one has to go back and forth to see what this was.

InterfacePython/TasmanianParticleSwarm.py
```diff
@@ -0,0 +1,295 @@
...
+        iNumDims = self.getNumDimensions()
+        aResult = np.zeros((iNumDims,), np.float64)
+        pLibDTSG.tsgParticleSwarmState_GetBestPosition(self.pStatePntr, aResult)
+        return aResult
+
+    def isPositionInitialized(self):
+        '''
+        Returns True if the particle positions have been initialized and False otherwise.
+        '''
+        return bool(pLibDTSG.tsgParticleSwarmState_IsPositionInitialized(self.pStatePntr))
```
mkstoyanov, line 161: Test this, `bool(0)` will be `False`, but I'm not sure about `bool(c_int(0))` which comes from C++.

If this causes issues, try
```
return (pLibDTSG.tsgParticleSwarmState_IsPositionInitialized(self.pStatePntr) != 0)
```

InterfacePython/TasmanianParticleSwarm.py
mkstoyanov: Change it to `self` here, the other one should also have `self` but the `.shape` can stay.

wwkong: Done.

InterfacePython/TasmanianParticleSwarm.py
wwkong: This was added to simplify the code in some of the functions, as we don't need to resize and call `as_ctypes()` on the NumPy arrays before passing them to the C wrappers.

I've changed them to `POINTER(c_double)` in the latest push.

InterfacePython/TasmanianParticleSwarm.py
wwkong: Changed.

InterfacePython/TasmanianParticleSwarm.py
wwkong: Both `bool(0)` and `bool(c_int(0))` return `False`, while `bool(1)` and `bool(c_int(1))` return `True` as expected.

I also tested these new functions in the sandbox, and they seem to work as expected.

mkstoyanov: OK, I wasn't sure since the documentation wasn't explicit. They talked about `int` but `c_int` can be argued as either int or not-int.

PR #678, opened 2022-05-11
added differentiation benchmark

mkstoyanov:
* add a benchmark for the differentiation option, available for all grid types
* example usage: `./SparseGrids/benchmarksgrid differentiate wavelet 6 1 2 level wavelet 1 10 10`
* see also `./SparseGrids/benchmarksgrid differentiate help`

### DIFF
--- a/SparseGrids/Benchmarks/benchDifferentiate.hpp
+++ b/SparseGrids/Benchmarks/benchDifferentiate.hpp
@@ -1,0 +1,45 @@
+#ifndef _TASMANIAN_BENCHMARK_DIFFERENTIATE_HPP
+#define _TASMANIAN_BENCHMARK_DIFFERENTIATE_HPP
+
+#include "benchCommon.hpp"
+
+bool benchmark_differentiate(std::deque<std::string> &args){
+    if (args.size() < 9) return false;
+
+    // report the test parameters to reference later
+    cout << "differentiate";
+    for(auto &s : args) cout << " " << s;
+
+    auto grid_family = getGridFamily(args);
+    if (grid_family == GridFamily::none) return false;
+
+    int num_dimensions, num_outputs, num_depth, order, iteratons, num_jumps;
+    TypeDepth dtype;
+    TypeOneDRule rule;
+
+    auto riter = readEntries(args.begin(), num_dimensions, num_outputs, num_depth, dtype, rule, order, iteratons, num_jumps);
+
+    auto extra = extractWeightsLimits(grid_family, num_dimensions, dtype, riter, args.end());
+
+    auto make_grid = getLambdaMakeGrid(grid_family, num_dimensions, num_outputs, num_depth, dtype, rule, order, extra);
+
+    num_jumps = std::max(num_jumps, 1); // make at least one test
+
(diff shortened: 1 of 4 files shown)


SparseGrids/Benchmarks/bench_main.cpp
```diff
@@ -24,14 +25,26 @@ int main(int argc, const char** argv){
...
+            pass = benchmark_evaluate(args, (test == bench_evaluate_mixed));
+            break;
+        case bench_differentiate:
+            pass = benchmark_differentiate(args);
+            break;
+        case bench_iweights:
+            pass = benchmark_iweights(args);
+            break;
+        default:
+            throw std::runtime_error("bench_main.cpp: invalid test type is switch statement!");
```
wwkong, line 46: Should the input be `"bench_main.cpp: invalid test type in switch statement!"`?

SparseGrids/Benchmarks/benchDifferentiate.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+    num_jumps = std::max(num_jumps, 1); // make at least one test
+
+    std::vector<std::vector<double>> inputs = getRandomVectors<double>(iteratons, num_dimensions, 1);
+
+    for(int jump = 0; jump < num_jumps; jump++){
+        auto grid = make_grid();
+        if (jump == 0) cout << " (points: " << grid.getNumPoints() << ")" << endl;
+        loadGenericModel(grid);
+        if (jump == 0) cout << "    note: reporting total time, does not normalize (divide) by iteratons\n";
```
wwkong, line 34: Make the time units here explicit, i.e.,
```
if (jump == 0) cout << "    note: reporting total time (ms), does not normalize (divide) by iteratons\n";
```

SparseGrids/Benchmarks/benchDifferentiate.hpp
```diff
@@ -0,0 +1,45 @@
+#ifndef _TASMANIAN_BENCHMARK_DIFFERENTIATE_HPP
+#define _TASMANIAN_BENCHMARK_DIFFERENTIATE_HPP
+
+#include "benchCommon.hpp"
+
+bool benchmark_differentiate(std::deque<std::string> &args){
+    if (args.size() < 9) return false;
```
wwkong, line 7: Currently, I'm allowed to call `benchmarksgrid differentiate` with >9 arguments (additional anisotropic weights and level limits). Why not have the function return `false` if we don't have exactly 9 arguments? i.e.,
```
if (args.size() != 9) return false;
```
Alternatively, we could add additional info to the help string.

SparseGrids/Benchmarks/bench_main.cpp
mkstoyanov: Yes!

SparseGrids/Benchmarks/benchDifferentiate.hpp
mkstoyanov: All time is in ms in all cases. Added the comment.

SparseGrids/Benchmarks/benchDifferentiate.hpp
mkstoyanov: Updated the text to indicate that we allow for extra parameters.

PR #682, opened 2022-05-18
Better particle swarm documentation + small changes

wwkong:


DREAM/Optimization/tsgParticleSwarm.hpp
```diff
@@ -50,26 +50,43 @@
...
+ * - getParticleVelocities(), setParticleVelocities()
+ * - getBestParticlePositions(), setBestParticlePositions()
+ * - getBestPosition()
+ * - clearBestParticles()
+ * - isPositionInitialized(), isVelocityInitialized(), isBestPositionInitialized()
+ *
+ * \par Initializing a Random Swarm
+ * Helper methods are available for initializing random particle positions and velocities.
+ * - initializeParticlesInsideBox()
+ *
```
mkstoyanov, line 83: Add another paragraph that explicitly lists the method to get the computed optimum, i.e., best-position. Also, give some explanation of what that means, e.g., the optimal value observed throughout the entire series across all particles.

DREAM/Optimization/tsgParticleSwarm.hpp
```diff
@@ -50,26 +50,43 @@
...
+ * \par Clearing the Cache
+ * When an optimization method is applied to the swarm, certain data related to the objective function are stored in cache variables
+ * inside this class. If an optimization method needs to be applied to the swarm with a different objective function than the one used
+ * to generate the cache, the cache should first be reset.
+ * - clearCache()
```
mkstoyanov, line 88: When you clear the cache and when you later make a new call to `ParticleSwarm`, do you first update and cache all particle positions and best positions?

DREAM/Optimization/tsgParticleSwarm.hpp
```diff
@@ -80,6 +97,15 @@ class ParticleSwarmState {
     //! \brief Constructor for a particle swarm state with the number of dimensions and the number of particles inferred from
     //! a set of input particle positions \b pp and a set of input particle velocities \b pv.
     ParticleSwarmState(const int num_dimensions, std::vector<double> &&pp, std::vector<double> &&pv);
+    //! \brief Copy constructor.
+    ParticleSwarmState(const ParticleSwarmState &source) = default;
+    //! \brief Move constructor.
+    ParticleSwarmState(ParticleSwarmState &&source) = default;
+    //! \brief Destructor.
+    ~ParticleSwarmState() = default;
```
mkstoyanov, line 105: You don't need to specify the default destructor, the compiler will do it automatically. You need the constructors, though.

DREAM/Optimization/tsgParticleSwarm.hpp
wwkong: Done. I've also made a separate paragraph for `getBestParticlePositions()` and its related functions.

DREAM/Optimization/tsgParticleSwarm.hpp
wwkong: I'm not sure I understand the question. 

If you are asking about what is being cached in `ParticleSwarm()` when the cache in the particle swarm state is uninitialized, then it is only the objection function values of the current/best particle positions and whether or not the current/best particles are inside the domain.

DREAM/Optimization/tsgParticleSwarm.hpp
wwkong: Changed. I removed the destructor.

DREAM/Optimization/tsgParticleSwarm.hpp
mkstoyanov: If you have two functionals, then the best position with respect to the first functional may not be the best with respect to the second functional. You may have to change the best position per particle and the total best position.

Not so much related to the PR, but it made me think.

PR #685, opened 2022-05-27
Add CLI and MATLAB wrappers for grid differentiation

wwkong:
I have tested it in MATLAB 202a on `titanmath` and it seems to be working as intended.

### DIFF
--- a/InterfaceMATLAB/tsgCoreTests.m
+++ b/InterfaceMATLAB/tsgCoreTests.m
@@ -1009,18 +1009,19 @@ for i=1:num_p
     vals(i,:) = f(p(i,:));
 end
 tsgLoadValues(lGrid, vals);
+% We must use MATLAB's deprecated seed setting system to ensure compatibility with Octave.
+rand('twister', 777);
 % Single point.
-rng(777);
 p0 = 2 * rand(1, 2) - 1.0;
 tsgDf0 = tsgDifferentiate(lGrid, p0);
-if (norm(tsgDf0 - Df(p0), 'F') > 1E-11)
+if (norm(tsgDf0 - Df(p0), 'fro') > 1E-11)
     error(['Mismatch in tsgDifferentiate: single point case']);
 end
 % Multiple points
 pk = 2 * rand(5, 2) - 1.0;
 tsgDfk =  tsgDifferentiate(lGrid, pk);
 for i=1:5
-    if (norm(tsgDfk{i} - Df(pk(i,:)), 'F') > 1E-11)
+    if (norm(tsgDfk{i} - Df(pk(i,:)), 'fro') > 1E-11)
         error(['Mismatch in tsgDifferentiate: multi point case']);
     end
 end
@@ -1037,18 +1038,19 @@ for i=1:num_p
     vals(i,:) = f(p(i,:));
 end
 tsgLoadValues(lGrid, vals);
+% We must use MATLAB's deprecated seed setting system to ensure compatibility with Octave.
+rand('twister', 777);
 % Single point.
-rng(777);
 p0 = 2 * rand(1, 2) - 1.0;
 tsgW0 = tsgGetDifferentiationWeights(lGrid, p0);
-if (norm(vals' * tsgW0 - Df(p0), 'F') > 1E-11)
+if (norm(vals' * tsgW0 - Df(p0), 'fro') > 1E-11)
     error(['Mismatch in tsgGetDifferentiationWeights: single point case']);
 end
 % Multiple points
 pk = 2 * rand(5, 2) - 1.0;
 tsgWk =  tsgGetDifferentiationWeights(lGrid, pk);
 for i=1:5
-    if (norm(vals' * tsgWk{i} - Df(pk(i,:)), 'F') > 1E-11)
+    if (norm(vals' * tsgWk{i} - Df(pk(i,:)), 'fro') > 1E-11)
         error(['Mismatch in tsgGetDifferentiationWeights: multi point case']);
     end
 end
--- a/InterfaceMATLAB/tsgCoreTests.m
+++ b/InterfaceMATLAB/tsgCoreTests.m
@@ -1009,19 +1009,17 @@ for i=1:num_p
     vals(i,:) = f(p(i,:));
 end
 tsgLoadValues(lGrid, vals);
-% We must use MATLAB's deprecated seed setting system to ensure compatibility with Octave.
-rand('twister', 777);
 % Single point.
 p0 = 2 * rand(1, 2) - 1.0;
 tsgDf0 = tsgDifferentiate(lGrid, p0);
-if (norm(tsgDf0 - Df(p0), 'fro') > 1E-11)
+if (norm(tsgDf0 - Df(p0)) > 1E-10)
     error(['Mismatch in tsgDifferentiate: single point case']);
 end
 % Multiple points
 pk = 2 * rand(5, 2) - 1.0;
 tsgDfk =  tsgDifferentiate(lGrid, pk);
 for i=1:5
-    if (norm(tsgDfk{i} - Df(pk(i,:)), 'fro') > 1E-11)
+    if (norm(tsgDfk{i} - Df(pk(i,:))) > 1E-10)
         error(['Mismatch in tsgDifferentiate: multi point case']);
     end
 end
@@ -1038,19 +1036,17 @@ for i=1:num_p
     vals(i,:) = f(p(i,:));
 end
 tsgLoadValues(lGrid, vals);
-% We must use MATLAB's deprecated seed setting system to ensure compatibility with Octave.
-rand('twister', 777);
 % Single point.
 p0 = 2 * rand(1, 2) - 1.0;
 tsgW0 = tsgGetDifferentiationWeights(lGrid, p0);
-if (norm(vals' * tsgW0 - Df(p0), 'fro') > 1E-11)
+if (norm(vals' * tsgW0 - Df(p0)) > 1E-10)
     error(['Mismatch in tsgGetDifferentiationWeights: single point case']);
 end
 % Multiple points
 pk = 2 * rand(5, 2) - 1.0;
 tsgWk =  tsgGetDifferentiationWeights(lGrid, pk);
 for i=1:5
-    if (norm(vals' * tsgWk{i} - Df(pk(i,:)), 'fro') > 1E-11)
+    if (norm(vals' * tsgWk{i} - Df(pk(i,:))) > 1E-10)
         error(['Mismatch in tsgGetDifferentiationWeights: multi point case']);
     end
 end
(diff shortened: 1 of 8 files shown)


Tasgrid/tasgridWrapper.cpp
```diff
@@ -600,6 +602,37 @@ bool TasgridWrapper::getIntegrate(){
...
+    if (x.empty()){
+        cerr << "ERROR: no points specified in " << xfilename << endl;
+        return false;
+    }
+    int num_in = grid.getNumDimensions();
+    int num_out = grid.getNumOutputs();
+    int num_points = x.getNumStrips();
+    std::vector<double> result(num_points * num_in * num_out);
+    std::vector<double> x_vec = x.release();
+    for (int i=0; i<num_points; i++)
```
mkstoyanov, line 628: Put an `#pragma omp parallel for` directive here.

Tasgrid/tasgridWrapper.hpp
```diff
@@ -56,6 +56,7 @@ enum TypeCommand{
 
     command_evaluate,
     command_integrate,
+    command_differentiate,
```
mkstoyanov, line 59: Good, but you should also add the derivative weights, e.g., similar to the get quadrature weights and get interpolation weights in matlab.

Tasgrid/tasgridWrapper.cpp
wwkong: Done.

Tasgrid/tasgridWrapper.hpp
wwkong: Done. 

Note that the interpretation of these weights can be a bit tricky (see the documentation in the new file `InterfaceMATLAB/tsgGetDifferentiationWeights.m`).

InterfaceMATLAB/tsgDifferentiate.m
```diff
@@ -0,0 +1,56 @@
+function [result] = tsgDifferentiate(lGrid, mX)
+%
+% [result] = tsgEvaluate(lGrid, points)
```
mkstoyanov, line 3: * which method is this? Correct with the above.

InterfaceMATLAB/tsgDifferentiate.m
```diff
@@ -0,0 +1,56 @@
...
+%
+% mX: an array of size [num_x, dimensions]
+%     specifies the points where the interpolant should be differentiated
+%     Note: do not confuse points here with the nodes of the grid
+%               here points are user specified points to differentiate the
+%               interpolant (or approximation)
+%
+% OUTPUT:
+%
+% result: an array of size [num_x, iOut * dimensions]
```
mkstoyanov, line 21: Ugh, this is now an ugly problem thanks to MATLAB.
* the output of differentiate is not a derivative or a gradient, but a Jacobian (multi-input/multi-output case)
* in C and C++, we use 1D arrays for matrices anyway, so we this is OK
* MATLAB and Python have native multi-d arrays, so we should respect that
* in Python, we do one derivative at a time, so we just return a matrix (right? we should)
* in MATLAB, there are no 3D arrays, but there are lists of matrixes
```
A = [1, 2; 3, 4]; B = [5, 6; 7, 8];
C = {A, B};
C[1]
C[2]
```
* basically, if `num_x == 1`, we should just return the 2D Jacobian, otherwise we should list of matrices as above.

InterfaceMATLAB/tsgGetDifferentiationWeights.m
```diff
@@ -0,0 +1,60 @@
...
+%
+% mX: an array of size [num_x, dimensions]
+%     specifies the points where the derivative should be evaluated
+%     Note: do not confuse points here with the nodes of the grid
+%           here points are user specified points to evaluate the
+%           derivative
+%
+% OUTPUT:
+%
+% weights: an array of size [num_x, number_of_points]
```
mkstoyanov, line 19: This is also incorrect, `weights` should hold `dimensions` somewhere in there.

Make it so that `f(points)' * weights` gives you the Jacobian in the case of `num_x == 1`. Then do the cell trick that I described above.

Tasgrid/tasgridWrapper.cpp
```diff
@@ -544,6 +547,29 @@ bool TasgridWrapper::getInterWeights(){
...
+    }
+    int num_w = x.getNumStrips();
+    int num_p = grid.getNumPoints() * grid.getNumDimensions();
+    Data2D<double> result(num_p, num_w);
+
+    #pragma omp parallel for
+    for(int i=0; i<num_w; i++)
+        grid.getDifferentiationWeights(x.getStrip(i), result.getStrip(i));
+
+    writeMatrix(outfilename, num_w, (int) num_p, result.getStrip(0));
```
mkstoyanov, line 568: * `num_p` is already `int`, no need to convert
* use `result.data()` it is more clear that you are dumping all data, as opposed to the first strip only

Tasgrid/tasgridWrapper.cpp
```diff
@@ -600,6 +626,38 @@ bool TasgridWrapper::getIntegrate(){
...
+    }
+    if (x.empty()){
+        cerr << "ERROR: no points specified in " << xfilename << endl;
+        return false;
+    }
+    int num_in = grid.getNumDimensions();
+    int num_out = grid.getNumOutputs();
+    int num_points = x.getNumStrips();
+    std::vector<double> result(num_points * num_in * num_out);
+    std::vector<double> x_vec = x.release();
```
mkstoyanov, line 651: Check `getDiffWeights()` and how much cleaner this becomes when you use `Data2D` as opposed to converting to a regular vector and then using complicated indexing.

mkstoyanov (changes requested):

InterfaceMATLAB/tsgDifferentiate.m
wwkong: Fixed.

InterfaceMATLAB/tsgDifferentiate.m
wwkong: Done.

Tasgrid/tasgridWrapper.cpp
wwkong: Fixed.

Tasgrid/tasgridWrapper.cpp
wwkong: Changed to use `Data2D` instead of `std::vector<double>`.

wwkong:
NOTE: I have also added some basic unit tests in `InterfaceMATLAB/tsgCoreTests.m` for `InterfaceMATLAB/tsgDifferentiate.m` and `InterfaceMATLAB/tsgGetDifferentiationWeights.m`.

wwkong:
NOTE: I have also added some basic unit tests in `InterfaceMATLAB/tsgCoreTests.m` for `InterfaceMATLAB/tsgDifferentiate.m` and `InterfaceMATLAB/tsgGetDifferentiationWeights.m`.

InterfaceMATLAB/tsgCoreTests.m
```diff
@@ -997,6 +997,62 @@ function tsgCoreTests()
 
 disp(['Utility functions:        PASS']);
 
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%%%                     tsgDifferentiate()                           %%%
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+[lGrid, p] = tsgMakeGlobal('_tsgcoretests_diff', 2, 3, 'leja', 'iptotal', 3);
+num_p = length(p);
+f = @(x) [x(1), x(1)*x(2)*x(2), x(2)*x(2)];
+Df = @(x) [1, 0; x(2)*x(2), 2*x(1)*x(2); 0, 2*x(2)];
```
mkstoyanov, line 1006: Can you check if this works on Octave. I suspect it does, but check anyway.

Tasgrid/tasgridWrapper.cpp
```diff
@@ -600,6 +626,37 @@ bool TasgridWrapper::getIntegrate(){
...
+    }
+    int num_points = x.getNumStrips();
+    int num_in = grid.getNumDimensions();
+    int num_out = grid.getNumOutputs();
+    Data2D<double> result(num_in * num_out, num_points);
+    #pragma omp parallel for
+    for (int i=0; i<num_points; i++)
+        grid.differentiate(x.getStrip(i), result.getStrip(i));
+
+    writeMatrix(outfilename, num_points, num_out * num_in, result.data());
```
mkstoyanov, line 655: The indexing is much better, but it is still prone to error (even in the rest of the code).

A better way to do it is to add overloads to `writeMatrix()` and `printMatrix()` that accept `Data2D<T>` and no sizes, then the overload will call the main method by taking the sizes from the number of strips and the size of the stride.

mkstoyanov: If you do this, you can also skip the declaration of the `int` variables, e.g.,
```
Data2D<double> result(grid.getNumDimensions() * grid.getNumOutputs(), x.getNumStrips());
for (int i=0; i<x.getNumStrips(); i++)  grid.differentiate(x.getStrip(i), result.getStrip(i));
writeMatrix(outfilename, result);
```

wwkong: Changed. I added the overloads and changed  `TasgridWrapper::getDifferentiate()` and `TasgridWrapper::getDiffWeights()` to use them.

InterfaceMATLAB/tsgCoreTests.m
mkstoyanov: What did you change specifically. If you must, don't use function handles but copy-paste the formulas. It may be a bad approach, but I don't want to drop support for either platform just because of two lines of code in testing.

Tasgrid/tasgridWrapper.hpp
```diff
@@ -181,7 +186,13 @@ class TasgridWrapper{
     void writeMatrix(std::string const &filename, int rows, int cols, const double mat[]) const;
     void printMatrix(int rows, int cols, const double mat[], bool isComplex = false) const;
 
-private:
+    // Overloads.
+    template<typename T>
+    inline void writeMatrix(std::string const &filename, Data2D<T> mat) { writeMatrix(filename, mat.getNumStrips(), mat.getStride(), mat.data()); }
```
mkstoyanov, line 191: Ouch, right in the performance!

Pass by reference and check const-correctness. Same for the other overload.

InterfaceMATLAB/tsgCoreTests.m
wwkong: The function handles were fine. The specific changes were

- `norm(x, 'F')` to `norm(x, 'fro')`
- `rng(seed)` to `rand('twister', seed)`

Only the latter is deprecated and not recommended by MATLAB.

mkstoyanov (approved):

PR #686, opened 2022-05-31
Add gradient descent algorithm and state object

wwkong:
- This PR currently includes some sandbox code for testing the features of the algorithm and state. We can remove this code when it is ready to merge.

### DIFF
--- a/DREAM/Optimization/tsgGradientDescent.cpp
+++ b/DREAM/Optimization/tsgGradientDescent.cpp
@@ -1,0 +1,103 @@
+
+/*
+ * Copyright (c) 2022, Miroslav Stoyanov & Weiwei Kong
+ *
+ * This file is part of
+ * Toolkit for Adaptive Stochastic Modeling And Non-Intrusive ApproximatioN: TASMANIAN
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following
+ * conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions
+ *    and the following disclaimer in the documentation and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse
+ *    or promote products derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ * IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+ * OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * UT-BATTELLE, LLC AND THE UNITED STATES GOVERNMENT MAKE NO REPRESENTATIONS AND DISCLAIM ALL WARRANTIES, BOTH EXPRESSED AND
+ * IMPLIED. THERE ARE NO EXPRESS OR IMPLIED WARRANTIES OF MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE, OR THAT THE USE OF
+ * THE SOFTWARE WILL NOT INFRINGE ANY PATENT, COPYRIGHT, TRADEMARK, OR OTHER PROPRIETARY RIGHTS, OR THAT THE SOFTWARE WILL
+ * ACCOMPLISH THE INTENDED RESULTS OR THAT THE SOFTWARE OR ITS USE WILL NOT RESULT IN INJURY OR DAMAGE. THE USER ASSUMES
+ * RESPONSIBILITY FOR ALL LIABILITIES, PENALTIES, FINES, CLAIMS, CAUSES OF ACTION, AND COSTS AND EXPENSES, CAUSED BY, RESULTING
+ * FROM OR ARISING OUT OF, IN WHOLE OR IN PART THE USE, STORAGE OR DISPOSAL OF THE SOFTWARE.
+ */
+
+#ifndef __TASMANIAN_GRADIENT_DESCENT_CPP
+#define __TASMANIAN_GRADIENT_DESCENT_CPP
+
+#include "tsgGradientDescent.hpp"
+
+namespace TasOptimization {
+
+GradientDescentState::GradientDescentState(const std::vector<double> x, const double ss, std::vector<double> lsc) :
+        num_dimensions((int) x.size()), stepsize(ss), candidate(x) {
+    if (line_search_coeffs.size() != 0 and line_search_coeffs.size() != 2)
+        throw std::runtime_error("ERROR: in GradientDescent(), expects line_search_coeffs.size() == 2 if non-empty");
+    line_search_coeffs = lsc;
+};
+
+void GradientDescent(const ObjectiveFunction f, const GradientFunction g, const ProjectionFunction proj, const int num_iterations,
+                     GradientDescentState &state) {
+
+    int num_dimensions = state.getNumDimensions();
+    std::vector<double> candidate = state.getCandidateRef();
+    std::vector<double> line_search_coeffs = state.getLineSearchCoeffs();
+
+    if (line_search_coeffs.empty()) {
+        // Constant stepsize scheme (no convergence guarantees).
+        std::vector<double> gradient;
+        for (int i=0; i<num_iterations; i++) {
+            g(candidate, gradient);
+            for (int j=0; j<num_dimensions; j++)
+                candidate[i] -= gradient[i] * state.stepsize;
+            proj(candidate, candidate);
+        }
+    } else {
+        // Variable stepsize scheme (convergence guarantee).
+        //
+        // Based on the paper:
+        //     Nesterov, Y. (2013). Gradient methods for minimizing composite functions. Mathematical programming, 140(1), 125-161.
+        std::vector<double> gradient, current_fval(1), next_candidate(num_dimensions), next_fval(1);
+        double next_stepsize = state.stepsize;
+        f(candidate, current_fval);
+        for (int i=0; i<num_iterations; i++) {
+            // Find the next stepsize/candidate point by making sure it satisfies the well-known descent inequality.
+            g(candidate, gradient);
+            double lhs = 0;
+            double rhs = 0;
+            next_stepsize *= line_search_coeffs[0]; // this is to avoid repeating the code for the first line search iteration
+            do {
+                next_stepsize /= line_search_coeffs[0];
+                for (int j=0; j<num_dimensions; j++)
+                    next_candidate[j] = candidate[j] - gradient[j] * next_stepsize;
+                proj(next_candidate, next_candidate);
+                f(next_candidate, next_fval);
+                lhs += next_fval[0] - next_fval[1];
+                for (int j=0; j<num_dimensions; j++) {
+                    double delta = next_candidate[j] - candidate[j];
+                    rhs += delta * delta / (2.0 * next_stepsize);
+                    lhs -= gradient[j] * delta;
+                }
+            } while (lhs > rhs);
+            // Update the key iterates manually.
+            std::swap(candidate, next_candidate);
+            std::swap(current_fval, next_fval);
+            next_stepsize *= line_search_coeffs[1];
+        }
+    }
+
+}
+
+}
+
+#endif
--- a/DREAM/Optimization/tsgGradientDescent.cpp
+++ b/DREAM/Optimization/tsgGradientDescent.cpp
@@ -40,7 +40,7 @@
 namespace TasOptimization {
 
 GradientDescentState::GradientDescentState(const std::vector<double> x, const double ss, std::vector<double> lsc) :
-        num_dimensions((int) x.size()), num_line_search_iterations(0), stepsize(ss), candidate(x) {
+        num_dimensions((int) x.size()), stepsize(ss), candidate(x) {
     if (line_search_coeffs.size() != 0 and line_search_coeffs.size() != 2)
         throw std::runtime_error("ERROR: in GradientDescent(), expects line_search_coeffs.size() == 2 if non-empty");
     line_search_coeffs = lsc;
@@ -49,9 +49,9 @@ GradientDescentState::GradientDescentState(const std::vector<double> x, const do
 void GradientDescent(const ObjectiveFunction f, const GradientFunction g, const ProjectionFunction proj, const int num_iterations,
                      GradientDescentState &state) {
 
-    int num_dimensions = state.getNumDimensions();
+    int num_dimensions = state.num_dimensions;
+    std::vector<double> line_search_coeffs = state.line_search_coeffs;
     std::vector<double> candidate = state.getCandidateRef();
-    std::vector<double> line_search_coeffs = state.getLineSearchCoeffs();
 
     if (line_search_coeffs.empty()) {
         // Constant stepsize scheme (does not have convergence guarantees).
@@ -69,41 +69,38 @@ void GradientDescent(const ObjectiveFunction f, const GradientFunction g, const 
         // Based on the paper:
         //     Nesterov, Y. (2013). Gradient methods for minimizing composite functions. Mathematical programming, 140(1), 125-161.
         std::vector<double> current_gradient(num_dimensions), current_fval(1), next_candidate(num_dimensions), next_fval(1);
-        double next_stepsize(state.stepsize), lhs, rhs;
-        int ls_iters = state.getNumLineSearchIterations();
-        for (int i=0; i<num_iterations; i++) {
-            // Optimistic stepsize update (see γ_d in the referenced paper above).
-            next_stepsize *= line_search_coeffs[1];
-            // Offset.
-            next_stepsize *= line_search_coeffs[0];
-            ls_iters--;
+        double stepsize(state.stepsize), current_iteration(0), lhs(0), rhs(0);
+        while(current_iteration < num_iterations) {
             // Find the next stepsize/candidate point by making sure it satisfies the well-known descent inequality (see
             // γ_u in the referenced paper above).
             f(candidate, current_fval);
             g(candidate, current_gradient);
             do {
-                ls_iters++;
+                if (current_iteration >= num_iterations) return;
                 lhs = 0;
                 rhs = 0;
-                next_stepsize /= line_search_coeffs[0];
                 for (int j=0; j<num_dimensions; j++)
-                    next_candidate[j] = candidate[j] - current_gradient[j] * next_stepsize;
+                    next_candidate[j] = candidate[j] - current_gradient[j] * stepsize;
                 proj(next_candidate, next_candidate);
                 f(next_candidate, next_fval);
                 lhs += next_fval[0] - current_fval[0];
                 for (int j=0; j<num_dimensions; j++) {
                     double delta = next_candidate[j] - candidate[j];
-                    rhs += delta * delta / (2.0 * next_stepsize);
+                    rhs += delta * delta / (2.0 * stepsize);
                     lhs -= current_gradient[j] * delta;
                 }
+                stepsize /= line_search_coeffs[0];
+                state.setStepsize(stepsize);
+                current_iteration++;
             } while (lhs > rhs + TasGrid::Maths::num_tol);
-            // Update the key iterates manually.
+            // Update the key iterates manually for the next loop.
             std::swap(candidate, next_candidate);
             std::swap(current_fval, next_fval);
+            state.setCandidate(candidate);
+            // Optimistic stepsize update (see γ_d in the referenced paper above).
+            stepsize *= line_search_coeffs[1] * line_search_coeffs[0];
+            state.setStepsize(stepsize);
         }
-        state.setNumLineSearchIterations(ls_iters);
-        state.setStepsize(next_stepsize);
-        state.setCandidate(candidate);
     }
 
 }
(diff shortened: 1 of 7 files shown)


DREAM/Optimization/tsgGradientDescent.hpp
```diff
@@ -0,0 +1,154 @@
...
+ * \par
+ * More information about the gradient descent algorithm can be found in the following paper:
+ *
+ * \par
+ * > Nesterov, Y. (2013). Gradient methods for minimizing composite functions. <em>Mathematical programming</em>, 140(1), 125-161.
+ *
+ */
+class GradientDescentState {
+  public:
+    //! \brief The default constructor is allowed.
```
mkstoyanov, line 88: * documentation says `allow`
* the next line says `delete`
* if all else fails, we can make a meme out of this one

mkstoyanov: On a related but tangent note, this is why "self-documented code" is a garbage notion.

Documentation gives you a layer of redundancy, if I see `GradientDescentState() = delete;` I know exactly what you are doing, but I have no way to check if this is what you wanted to do. When you have the code and the documentation, you double-down and affirm the correct thing (e.g., "yes, I want to delete this") or you make an error that we can now correct. Without documentation, an error in `default/delete` is much harder or catch.

DREAM/Optimization/tsgGradientDescent.hpp
```diff
@@ -0,0 +1,154 @@
...
+    GradientDescentState() = delete;
+    //! \brief Constructor for a gradient descent state with the initial candidate and stepsize.
+    GradientDescentState(const std::vector<double> &x, const double stepsize);
```
mkstoyanov, line 91: I don't think the `stepsize` belongs in the state. I think this is a parameter that goes into the call to the descent method. It is common for descent methods to vary the `stepsize` or to adjust every few iterations. Even without an automated algorithm to adjust that, we should allow the user to make the adjustment.

mkstoyanov: Also, why call it `x` when it should be `initial_iterate` or something related.

DREAM/Optimization/tsgGradientDescent.hpp
```diff
@@ -0,0 +1,154 @@
...
+    //! \brief Move assignment.
+    GradientDescentState& operator=(GradientDescentState &&source) = default;
+
+    //! \brief Return the number of dimensions.
+    inline int getNumDimensions() const {return num_dimensions;}
+    //! \brief Return the stepsize.
+    inline double getStepsize() const {return stepsize;}
+    //! \brief Return the current candidate point.
+    inline void getCandidate(double x[]) const {std::copy_n(candidate.begin(), num_dimensions, x);}
+    inline std::vector<double> getCandidate() const {return candidate;}
```
mkstoyanov, line 106: missing `//! \brief`

DREAM/Optimization/tsgGradientDescent.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+    //! \brief Set the stepsize.
+    inline void setStepsize(const double ss) {stepsize = ss;}
+    //! \brief Set the current candidate point.
+    inline void setCandidate(const double x[]) {std::copy_n(x, num_dimensions, candidate.begin());}
+    inline void setCandidate(const std::vector<double> &x) {
```
mkstoyanov, line 112: Looking at the state as a whole, we can probably implement the whole thing with:
```
using GradientDescentState = std::vector<double>;
```

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,104 @@
...
+                     GradientDescentState &state, const std::vector<double> &line_search_coeffs) {
+
+    if (line_search_coeffs.size() != 0 and line_search_coeffs.size() != 2)
+        throw std::runtime_error("ERROR: in GradientDescent(), expects line_search_coeffs.size() == 2 if non-empty");
+
+    int num_dimensions = state.num_dimensions;
+    std::vector<double> &candidate = state.getCandidateRef();
+
+    if (line_search_coeffs.empty()) {
+        // Constant stepsize scheme (does not have convergence guarantees).
```
mkstoyanov, line 54: I disagree. We need to talk about this one.

DREAM/Optimization/tsgGradientDescent.hpp
wwkong: Fixed.

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,126 @@
...
+            if (status.num_iterations >= num_iterations)
+                break;
+            for (int j=0; j<num_dimensions; j++)
+                x[j] = x_prev[j] - g_at_x_prev[j] * state.adaptive_stepsize;
+            x = proj(x);
+            f_at_x = f(x);
+            lhs += f_at_x - f_at_x_prev;
+            for (int j=0; j<num_dimensions; j++) {
+                double delta = x[j] - x_prev[j];
+                rhs += delta * delta / (2.0 * state.adaptive_stepsize);
```
mkstoyanov, line 75: This saves one flop, how much does this depend on the projection of `x`, can we do it here?
```
double ngrad = g_at_x_prev[j] * g_at_x_prev[j] * state.adaptive_stepsize;
rhs += 0.5 * ngrad;
lhs -= ngrad;
```

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,126 @@
...
+    // Wrapper to the proximal version with projection function == identity function.
+    return GradientDescent(f, g, identity, state, line_search_coeffs, num_iterations, tolerance);
+};
+
+OptimizationStatus GradientDescent(const GradientFunctionSingle &g, std::vector<double> &x, const double stepsize, const int num_iterations,
+                                   const double tolerance) {
+    OptimizationStatus status;
+    status.num_iterations = 0;
+    status.stationarity_residual = std::numeric_limits<double>::max();
+    std::vector<double> x_prev(x), g_at_x_prev(g(x_prev)), g_at_x(x.size());
```
mkstoyanov, line 109: Naming convention? I don't like the `_at_` portion of the name, `g_at_x` has 5 characters, but only `g` and `x` have a meaning, the mid part is just clutter to differentiate from the name of the input variable.

I know you need to differentiate between `g_at_x` and the GradientFunctionSingle `g`, but how about calling things:
```
OptimizationStatus GradientDescent(const GradientFunctionSingle &grad, std::vector<double> &state, ...) {}
```
Then you can have variables like `x`, `gx` and `gstate`, where `x` is the internal temp and `state` is the result. The variables are easy to differentiate and the inputs are more expressive.

Similarly `f` can be `objective` or `cost` or `func` (for functional)

DREAM/Optimization/tsgOptimizationUtils.hpp
```diff
@@ -57,6 +57,16 @@
 
 namespace TasOptimization {
 
+/*! \internal
+ * \ingroup OptimizationUtil
+ *
+ * Stores information about the run of an optimization algorithm.
+ */
+struct OptimizationStatus {
+    int num_iterations;
```
mkstoyanov, line 66: Is this the max or the actual iterations?

Use better names:
* performed_iterations or done_iterations
* max_iterations

DREAM/Optimization/tsgOptimizationUtils.hpp
```diff
@@ -57,6 +57,16 @@
...
+    double stationarity_residual;
```
mkstoyanov, line 67: Change this to `residual` and we will explain what it means for each method.

Here we can add a note that the definition of the `residual` can depend on the algorithm.

DREAM/Optimization/tsgOptimizationUtils.hpp
```diff
@@ -105,6 +118,45 @@ inline ObjectiveFunction makeObjectiveFunction(const int num_dimensions, const O
     };
 }
 
+/*! \ingroup OptimizationUtil
+ * \brief Generic non-batched gradient function signature.
+ *
+ * Accepts a single input \b x and returns the gradient at the point \b x.
+ */
+using GradientFunctionSingle = std::function<std::vector<double>(const std::vector<double> &x_batch)>;
```
mkstoyanov, line 126: How does this work in DREAM? Do we return a vector or do we pass in a pre-allocated vector or pointer, which avoids the allocation.

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+#include "tsgGradientDescent.hpp"
+
+namespace TasOptimization {
+
+GradientDescentState::GradientDescentState(const std::vector<double> &x0, const double lambda0) :
+        num_dimensions((int) x0.size()), adaptive_stepsize(lambda0), x(x0) {};
+
+OptimizationStatus GradientDescent(const ObjectiveFunctionSingle &f, const GradientFunctionSingle &g, const ProjectionFunctionSingle &proj,
+                                   GradientDescentState &state, const std::vector<double> &line_search_coeffs, const int num_iterations,
```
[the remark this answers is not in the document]
wwkong: Changed. This has been split into two parameters `increase_coeff` and `decrease_coeff`.

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,126 @@
...
+    OptimizationStatus status;
+    status.num_iterations = 0;
+    status.stationarity_residual = std::numeric_limits<double>::max();
+    std::vector<double> x_prev(x), g_at_x_prev(g(x_prev)), g_at_x(x.size());
+    for (int i=0; i<num_iterations; i++) {
+        for (size_t j=0; j<x.size(); j++)
+            x[j] -= g_at_x_prev[j] * stepsize;
+        g_at_x = g(x);
+        status.num_iterations++;
+        status.stationarity_residual = compute_stationarity_residual(x, x_prev, g_at_x, g_at_x_prev, stepsize);
```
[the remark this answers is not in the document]
wwkong: Fixed the calculation and the `compute_stationary_residual()` function. The residual should just be
```
status.residual = std::sqrt(g_at_x * g_at_x)
```

DREAM/Optimization/tsgGradientDescent.cpp
wwkong: Changed. `f` is now `func`, `g` is now `grad`, and the `_at_` portion has been removed in all variables.

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,126 @@
...
+    
+}
+
+OptimizationStatus GradientDescent(const ObjectiveFunctionSingle &f, const GradientFunctionSingle &g, GradientDescentState &state,
+                                   const std::vector<double> &line_search_coeffs,  const int num_iterations, const double tolerance) {
+    // Wrapper to the proximal version with projection function == identity function.
+    return GradientDescent(f, g, identity, state, line_search_coeffs, num_iterations, tolerance);
+};
+
+OptimizationStatus GradientDescent(const GradientFunctionSingle &g, std::vector<double> &x, const double stepsize, const int num_iterations,
```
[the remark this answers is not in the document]
wwkong: Changed. I will need to make a separate PR to do the same in the particle swarm algorithm.

DREAM/Optimization/tsgOptimizationUtils.hpp
wwkong: This is the actual iterations, so I have renamed it to `performed_iterations`.

DREAM/Optimization/tsgOptimizationUtils.hpp
wwkong: In DREAM, we pass a pre-allocated vector to the function. Should we do this here as well? I assumed RVO would help avoid extra allocations.

mkstoyanov: It's better to be consistent and pass in a pre-allocated vector.

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,130 @@
...
+ */
+
+#ifndef __TASMANIAN_GRADIENT_DESCENT_CPP
+#define __TASMANIAN_GRADIENT_DESCENT_CPP
+
+#include "tsgGradientDescent.hpp"
+
+namespace TasOptimization {
+
+GradientDescentState::GradientDescentState(const std::vector<double> &x0, const double lambda0) :
```
mkstoyanov, line 41: Small things like that can go in the header, may help with inlining.

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,130 @@
...
+                xStep[j] = x0[j] - gx0[j] * state.adaptive_stepsize;
+            xStep = proj(xStep);
+            fxStep = func(xStep);
+            lhs += fxStep - fx0;
+            for (int j=0; j<num_dimensions; j++) {
+                double delta = xStep[j] - x0[j];
+                rhs += delta * delta / (2.0 * state.adaptive_stepsize);
+                lhs -= gx0[j] * delta;
+            }
+            state.adaptive_stepsize /= decrease_coeff;
```
mkstoyanov, line 76: If `lhs < rhs` do you still reduce the `adaptive_stepsize`?

DREAM/Optimization/tsgGradientDescent.hpp
```diff
@@ -0,0 +1,204 @@
...
+    GradientDescentState() = delete;
+    //! \brief Constructor for a gradient descent state with the initial candidate and stepsize.
+    GradientDescentState(const std::vector<double> &x0, const double lambda0);
+    //! \brief Copy constructor.
+    GradientDescentState(const GradientDescentState &source) = default;
+    //! \brief Move constructor.
+    GradientDescentState(GradientDescentState &&source) = default;
+
+    //! \brief Move assignment.
+    GradientDescentState& operator=(GradientDescentState &&source) = default;
```
mkstoyanov, line 98: How about copy-assignment, we should allow that one too.

DREAM/Optimization/tsgOptimizationUtils.hpp
wwkong: Done. I changed the signature of `GradientFunctionSingle` and `ProjectionFunctionSingle` to have a pre-allocated vector.

DREAM/Optimization/tsgGradientDescent.cpp
wwkong: Changed. I moved this constructor to the header file.

DREAM/Optimization/tsgGradientDescent.cpp
wwkong: Yes. This is okay because the stepsize is then multiplied by `decrease_coeff` right after the termination of the while-loop.

To make the logic of this approach more clear, it's helpful to consider all the cases that occur and what we expect the adaptive stepsize to be at the end of one iteration of the main while loop. 

Specifically, denoting `stepsize0` to be stepsize at the beginning of an iteration of the `while(status.performed_iterations < num_iterations)` loop, we expect:

**Case 1 - A break occurs inside the do-while loop using the `(lhs > rhs)` condition**
`adaptive_stepsize = stepsize0 / (decrease_coeff ^ performed_iterations)`

**Case 2 - A break occurs due to the `status.residual <= tolerance` condition**
`adaptive_stepsize = stepsize0 / (decrease_coeff ^ [performed_iterations - 1])`

**Case 3 - The optimistic update is performed**
`adaptive_stepsize = stepsize0 / (decrease_coeff ^ [performed_iterations - 1] * increase_coeff)`

We can discuss more if this is confusing.

DREAM/Optimization/tsgGradientDescent.hpp
wwkong: Sure. I have added copy assignment to the header.

wwkong:
**Note**: I changed the location of the iteration swaps, e.g., `swap(x0, x)`, due to the change in function signatures in `GradientFunctionSingle` and `ProjectionFunctionSingle`.

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+OptimizationStatus GradientDescent(const GradientFunctionSingle &grad, const double stepsize, const int num_iterations,
+                                   const double tolerance, std::vector<double> &state) {
+    OptimizationStatus status;
+    status.performed_iterations = 0;
+    status.residual = std::numeric_limits<double>::max();
+    std::vector<double> x0(state), gx0(state.size()), gx(state.size());
+    grad(state, gx);
+
+    for (int i=0; i<num_iterations; i++) {
```
mkstoyanov, line 109: This is misleading, the main use case is to run until the residual drops below the tolerance and the special case is to use the `num_iterations` to break early. Use a while loop with both conditions, so it is clear.

```
while (status.residual > tolerance and status.performed_iterations < max_iterations){
    status.performed_iterations++;
    ....
    status.residual = ....
}

return status;
```

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,128 @@
...
+                                   const double tolerance, GradientDescentState &state) {
+    // Wrapper to the proximal version with projection function == identity function.
+    return GradientDescent(func, grad, identity, increase_coeff, decrease_coeff, num_iterations, tolerance, state);
+};
+
+OptimizationStatus GradientDescent(const GradientFunctionSingle &grad, const double stepsize, const int num_iterations,
+                                   const double tolerance, std::vector<double> &state) {
+    OptimizationStatus status;
+    status.performed_iterations = 0;
+    status.residual = std::numeric_limits<double>::max();
```
mkstoyanov, line 105: I personally feel very uneasy when I operate so close to numeric instability, and in the current implementation (and the one that I propose below), this isn't necessary.
* currently, you're not doing any residual testing until you overwrite this, you're not using this value so it can be anything
* in the proposed (and current) solution, you can set:
```
status.residual = tolerance + 1.0;
```

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,128 @@
...
+OptimizationStatus GradientDescent(const ObjectiveFunctionSingle &func, const GradientFunctionSingle &grad,
+                                   const double increase_coeff, const double decrease_coeff, const int num_iterations,
+                                   const double tolerance, GradientDescentState &state) {
+    // Wrapper to the proximal version with projection function == identity function.
+    return GradientDescent(func, grad, identity, increase_coeff, decrease_coeff, num_iterations, tolerance, state);
+};
+
+OptimizationStatus GradientDescent(const GradientFunctionSingle &grad, const double stepsize, const int num_iterations,
+                                   const double tolerance, std::vector<double> &state) {
+    OptimizationStatus status;
```
mkstoyanov, line 103: You can initialize both performed iterations and residual right here:
```
OptimizationStatus status{0, tolerance + 1.0}; // {performed iterations, residual}
```

DREAM/Optimization/tsgGradientDescent.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+OptimizationStatus GradientDescent(const ObjectiveFunctionSingle &func, const GradientFunctionSingle &grad,
+                                   const double increase_coeff, const double decrease_coeff, const int num_iterations,
+                                   const double tolerance, GradientDescentState &state) {
+    // Wrapper to the proximal version with projection function == identity function.
+    return GradientDescent(func, grad, identity, increase_coeff, decrease_coeff, num_iterations, tolerance, state);
+};
+
+OptimizationStatus GradientDescent(const GradientFunctionSingle &grad, const double stepsize, const int num_iterations,
```
mkstoyanov, line 101: Call the variable `max_iterations` since it is a fail-safe mechanism and not the actual number of iterations that you want to perform.

DREAM/Optimization/tsgOptimizationUtils.hpp
```diff
@@ -71,6 +81,9 @@ inline void checkVarSize(const std::string method_name, const std::string var_na
     }
 }
 
+// Functions used in optimization.
+using GenericBatchedFunction = std::function<void(const std::vector<double> &x_batch, std::vector<double> &y_batch)>;
```
mkstoyanov, line 85: Multiple functions have the same signature, but setting them all the the same `Generic` thing isn't helpful.
* the user has to look up the generic function to see what's going
* the user has to then read the documentation of the specific instance
* sometime `x_batch` is not actually a batch ... confusing

Here is a better approach:
* remove the generic function
* repeat the signature every time, but have different names for `x` and `y` as appropriate
```
ObjectiveFunction = std::function<void(const std::vector<double> &x_batch, std::vector<double> &fval_batch)>;
GradientFunctionSingle = std::function<void(const std::vector<double> &x_single, std::vector<double> &grad)>;
ProjectionFunctionSingle = std::function<void(const std::vector<double> &x_single, std::vector<double> &proj)>;
```
Have the documentation too.

DREAM/Optimization/tsgGradientDescent.hpp
```diff
@@ -0,0 +1,208 @@
...
+                                              const double tolerance, GradientDescentState &state);
+
+  protected:
+    #ifndef __TASMANIAN_DOXYGEN_SKIP_INTERNAL
+    //! \brief Returns a reference to the current point.
+    inline std::vector<double> &getXRef() {return x;}
+    #endif
+
+  private:
+    int num_dimensions;
```
mkstoyanov, line 175: This variable is redundant as it always needs to be equal to `x.size()`

DREAM/Optimization/tsgGradientDescent.cpp
wwkong: Changed using the suggestion below this one.

DREAM/Optimization/tsgGradientDescent.cpp
wwkong: Changed.

DREAM/Optimization/tsgOptimizationUtils.hpp
wwkong: Changed. I have used the suggestions above.

DREAM/Optimization/tsgGradientDescent.hpp
```diff
@@ -0,0 +1,208 @@
...
+     * \param state Holds the state of the gradient descent algorithm, see TasOptimization::GradientDescentState
+     */
+    friend OptimizationStatus GradientDescent(const ObjectiveFunctionSingle &f, const GradientFunctionSingle &g,
+                                              const double increase_coeff, const double decrease_coeff, const int num_iterations,
+                                              const double tolerance, GradientDescentState &state);
+
+  protected:
+    #ifndef __TASMANIAN_DOXYGEN_SKIP_INTERNAL
+    //! \brief Returns a reference to the current point.
+    inline std::vector<double> &getXRef() {return x;}
```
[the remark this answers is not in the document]
wwkong: Changed. The signatures in the various `GradientDescent` functions all have `GradientDescentState &state` instead of `std::vector<double> &state`, and I have removed the `GradientDescentState::getXRef()` method.

DREAM/Optimization/tsgGradientDescent.hpp
wwkong: Changed. `num_dimensions` has been removed.

DREAM/tasdreamExternalTests.cpp
```diff
@@ -566,6 +566,41 @@ bool DreamExternalTester::performTests(TypeDREAMTest test){
 void testDebug(){
     cout << "Debug Test" << endl;
     cout << "Put here testing code and call this with ./dreamtest debug" << endl;
+
```
mkstoyanov, line 569: If the code is ready, move this to a separate test or remove altogether.

Note if this is complete.

wwkong: I've removed this code and will set up a separate PR for testing.

The code should be complete and ready for merging now.

PR #687, opened 2022-06-14
Add Python ParticleSwarm tests + small bug fixes

wwkong:


InterfacePython/testOptimization.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertFalse(state.isPositionInitialized())
+        self.assertFalse(state.isVelocityInitialized())
+        self.assertFalse(state.isBestPositionInitialized())
+        self.assertFalse(state.isCacheInitialized())
+        # Loading/Unloading tests.
+        pp = np.random.rand(iNumParticles, iNumDimensions)
+        state.setParticlePositions(pp)
+        self.assertTrue(state.isPositionInitialized())
+        self.assertTrue(np.array_equal(pp, state.getParticlePositions()))
+        pv = np.random.rand(iNumParticles, iNumDimensions)
```
mkstoyanov, line 68: I would put something deterministic here, like 1, 2, 3, 4 ..., that way you can pinpoint the problem more easily, especially if it comes down to indexing.

In general, I don't like testing with random data, as it adds another potential point of failure, variability in tolerances, etc. Fixing the random seed helps, but not always. This test can easily be deterministic.

InterfacePython/testOptimization.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertEqual(iNumParticles, state.getNumParticles())
+        self.assertFalse(state.isPositionInitialized())
+        self.assertFalse(state.isVelocityInitialized())
+        self.assertFalse(state.isBestPositionInitialized())
+        self.assertFalse(state.isCacheInitialized())
+        # Loading/Unloading tests.
+        pp = np.random.rand(iNumParticles, iNumDimensions)
+        state.setParticlePositions(pp)
+        self.assertTrue(state.isPositionInitialized())
+        self.assertTrue(np.array_equal(pp, state.getParticlePositions()))
```
mkstoyanov, line 67: There is a way to do this with one command, like assert almost equal, look for it among the other tests. It also works with given tolerance (could be zero).

InterfacePython/testOptimization.py
```diff
@@ -0,0 +1,112 @@
...
+        f = lambda x_batch : np.apply_along_axis(shc, 1, x_batch)
+        inside = lambda x : bool((-3 <= x[0]) and (x[0] <= 3) and (-2 <= x[1]) and (x[1] <= 2))
+        # Main call + tests.
+        Opt.ParticleSwarm(f, 1, inside, state, 0.5, 2, 2)
+        self.assertTrue(state.isCacheInitialized())
+        state.clearCache()
+        self.assertFalse(state.isCacheInitialized())
+        iNumIterations = 200
+        Opt.ParticleSwarm(f, iNumIterations, inside, state, 0.5, 2, 2)
+        self.assertTrue(np.allclose(np.array([-0.08984201368301331, +0.7126564032704135]), state.getBestPosition()) or
```
mkstoyanov, line 103: Can adjust to almost equal here.

InterfacePython/testOptimization.py
```diff
@@ -0,0 +1,112 @@
...
+        shc = lambda x : (4 - 2.1*x[0]*x[0] + x[0]*x[0]*x[0]*x[0]/3)*x[0]*x[0] + x[0]*x[1] + (-4.0 + 4.0*x[1]*x[1])*x[1]*x[1]
+        f = lambda x_batch : np.apply_along_axis(shc, 1, x_batch)
+        inside = lambda x : bool((-3 <= x[0]) and (x[0] <= 3) and (-2 <= x[1]) and (x[1] <= 2))
+        # Main call + tests.
+        Opt.ParticleSwarm(f, 1, inside, state, 0.5, 2, 2)
+        self.assertTrue(state.isCacheInitialized())
+        state.clearCache()
+        self.assertFalse(state.isCacheInitialized())
+        iNumIterations = 200
+        Opt.ParticleSwarm(f, iNumIterations, inside, state, 0.5, 2, 2)
```
mkstoyanov, line 102: How do you control for randomness here. We must fix the random seed, otherwise this will fail periodically.

InterfacePython/testOptimization.py
wwkong: Changed. I used `numpy.arange()` to fill in the values.

InterfacePython/testOptimization.py
wwkong: Changed to use `numpy.testing.assert_array_equal()`.

InterfacePython/testOptimization.py
wwkong: There are two global minima in the six-hump-camel function, and we don't know which one the particle swarm method will converge to. So, we should check both points and avoid throwing an exception if `state.getBestPosition()` is at one of them, which is what the code is doing.

PR #723, opened 2023-02-09
added overloads for string

mkstoyanov:
* add overloads to read directly from strings


mkstoyanov:
* add overloads to read directly from strings

PR #753, opened 2024-02-15
Improve pip options

mkstoyanov:
* added option to disable recommendeds in the pip installer


mkstoyanov:
* improve pip options

PR #773, opened 2024-06-12
Fix missing cmethod

mkstoyanov:
Add the missing C methods from the C header.

@MichelJuillard Thanks for catching those.


MichelJuillard:
Thanke @mkstoyanov for the superfast reaction :)

